PR #20, opened 2024-03-04
make typescript compile

michaelglass:
npx tsc now compiles the chat widget

### DIFF
--- a/packages/chat-widget/tsconfig.json
+++ b/packages/chat-widget/tsconfig.json
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "target": "es6",
     "module": "esnext",
+    "moduleResolution": "bundler",
     "lib": ["dom"],
     "jsx": "react",
     "skipLibCheck": true,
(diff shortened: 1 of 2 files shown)


peterszerzo:
@michaelglass there is indeed something strange going on here that we should look into, what strikes me off the bat is that the regular `tsc` binary is suggesting fixes that don't look right (`Theme` from `@emotion/react` when what we intend to use is the local custom `Theme`). Conversely, if I make what I think is a legitimate type error, `rollup` alerts me as it should.

Can we move this PR into draft?

packages/chat-widget/tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "compilerOptions": {
     "target": "es6",
-    "module": "esnext",
+    "module": "nodenext",
```
michaelglass, line 4: this is the meat of this change. Unclear to me why targeting `esnext` won't pull in `types.d.ts` but it apparently doesn't :-/

peterszerzo (approved):

peterszerzo (approved):

PR #22, opened 2024-03-04
voice compass docs

michaelglass:
this is ready for review but awaiting some documentation links from nlx journey docs.


packages/website/package.json
```diff
@@ -19,6 +19,7 @@
     "react-dom": "^18.2.0",
     "react-markdown": "^9.0.0",
     "react-router-dom": "^6.17.0",
+    "remark-gfm": "^4.0.0",
```
michaelglass, line 22: peter & I discussed initial documentation, and the use case for this was removed. 

I think it's still worth keeping because I can imagine tables being useful. But we can also get rid of it until we actually use the functionality.

packages/website/src/components/ChatConfiguration.tsx
```diff
@@ -188,38 +189,6 @@ export const ThemeEditor: FC<{
   );
 };
 
-const RadioList = <T extends unknown>({
```
michaelglass, line 191: extracted to its own component. Peter wanted me to use this component for voice compass examples

packages/website/src/content/06-01-multimodal-getting-started.tsx
```diff
@@ -19,6 +33,14 @@ In a bundled JavaScript application or Node.js:
 ~~~js
 ${voiceCompassSetupSnippet({ environment: Environment.Bundle })}
 ~~~
+
+[flow]: https://docs.studio.nlx.ai/intentflows/overview
+
+[//]: <> (TODO: add proper links...)
```
michaelglass, line 39: these are ~ blocking merge

peterszerzo (dismissed):

peterszerzo (approved):

PR #23, opened 2024-03-04
small voice compass changes

michaelglass:
a handful of small changes to the voice compass API. Want to sync with @sam-trost about these

These are all good, but talking to Sam, I'd like to try scrapping most of this API and making it much much simpler. 

from my understanding: Voice Compass used to model a state machine internally, but now requires the user to model the state machine. Considering that: I think the small complexity around state management can go away and can be modeled in the documentation. So while this PR is probably a bunch of baby steps in the right direction, the subsequent PR will try trimming the library down to the bare minimum

### DIFF
--- a/packages/voice-compass/src/index.ts
+++ b/packages/voice-compass/src/index.ts
@@ -96,7 +96,7 @@ export const create = (config: Config): VoiceCompass => {
       })
       .catch((err: Error) => {
         if (config.debug) {
-          console.error(`× step: ${payload.stepId}`, err);
+          console.error(`× step: ${payload.stepId}`, err, payload);
         }
         return {
           error: `Something went wrong`,
--- a/packages/voice-compass/src/index.ts
+++ b/packages/voice-compass/src/index.ts
@@ -23,10 +23,21 @@ export interface Config {
 
 export interface StepData {
   stepId: string;
-  context?: Record<string, any>;
+  context?: Context;
 }
 
-export type Context = Record<string, any>;
+/** A JSON stringify-able value. */
+export type jsonable =
+  | string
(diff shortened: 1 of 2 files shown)


packages/voice-compass/src/index.ts
```diff
@@ -129,8 +125,7 @@ export const create = (config: Config): VoiceCompass => {
       });
     }
     lastUpdate = {
-      // TODO: sort out whether optional stepID's are even allowed
```
michaelglass, line 132: @sam-trost is a step id optional?

michaelglass: from sam: confirmed

packages/voice-compass/src/index.ts
```diff
@@ -23,10 +23,21 @@ export interface Config {
...
+  | string
+  | number
+  | boolean
+  | null
+  | undefined
+  | jsonable[]
+  | { [key: string]: jsonable }
+  | { toJSON(): jsonable };
+
+export type Context = Record<string, jsonable>;
```
jakub-nlx, line 40: I'm not convinced this is really an improvement.

michaelglass: mostly: functions

peterszerzo: I would stick to the `any` here, pretty sure some user will have a not-quite-so-perfectly typed codebase and the type definition will cause them trouble.

The other interesting avenue would be to add `context` as a kind of type parameter/generic to the client somehow, because this is code we can generate for download from the platform.

jakub-nlx (approved):

PR #31, opened 2024-03-17
generate api docs from tsdoc

michaelglass:
this is a prototype of generating documentation from tsdoc.

On the one hand, it's a little more verbose than the most basic of docs.

On the other hand it's more consistent and probably easier to keep in sync.

Maybe there's a partially-manual version where the generated markdown is simplified before consumed by the docs.

this builds on top of #22, the base branch is set to that for easier review.

Another option would be having a third API reference website that's more purely tsdocs generated from code. Would help to dis-entangle deploys and styles from the docs and API pages


packages/website/src/content/06-03-multimodal-api-reference.md
```diff
@@ -0,0 +1,230 @@
+
```
michaelglass, line 1: this file is generated from `npm run publish-docs`. I also removed a single superfluous doc line

packages/website/src/main.tsx
```diff
@@ -35,7 +35,7 @@ const App: FC<{}> = () => {
 };
 
 createRoot(document.getElementById("app") as Element).render(
-  <HashRouter>
+  <BrowserRouter>
```
michaelglass, line 38: hash router means: anchor tags (used in the markdown) don't work

packages/voice-compass/typedoc.cjs
```diff
@@ -0,0 +1,12 @@
+/** @type {import('typedoc').TypeDocOptions} */
+module.exports = {
+  sort: ["source-order", "kind", "instance-first", "alphabetical"],
+  entryPoints: ["./src/index.ts"],
+  out: "docs",
+  validation: { notExported: true, invalidLink: true, notDocumented: true },
+  treatWarningsAsErrors: true,
+  treatValidationWarningsAsErrors: true,
+  categoryOrder: ["Setup", "Client"],
+  plugin: ["typedoc-plugin-markdown"],
```
michaelglass, line 10: the markdown plugin is a proxy for a minimally-styled version of typedoc's output.

for more examples of how typedoc usually looks in the wild:

https://typedoc.org/api/
https://typedoc.org/example/

(or with a non-standard theme:)
https://v1.baklava.tech/api/index.html#/@baklavajs%2Fcore

packages/website/src/main.tsx
peterszerzo: @michaelglass thanks so much for changing the router, hash routing was set up at a time when we were still on github pages and it was easier not to have to deal with a base path different from `/`, this is better.

Please keep in mind that explicit references to hash-based URL's are present throughout the app (next page / prev page navigation for example), let's make sure these are all compatible with the new routing.

peterszerzo (approved):

PR #35, opened 2024-03-27
fix CI and some minor cleanups along the way

michaelglass:
fixes CI -- and did some cleanups while I was trying to diagnose the issue


packages/chat-core/tsconfig.json
```diff
@@ -1,14 +1,14 @@
 {
   "compilerOptions": {
-    "target": "es6",
```
michaelglass, line 3: I believe these are developer only because rollup handles its own tsconfig

PR #37, opened 2024-03-28
support browser router on github pages

michaelglass:
mislead by dev environment differing from prod environment.

Github Actions doesn't rewrite URLs on refresh, so using BrowserRouter meant loading, e.g. `/foo` would just 404.

This PR
- adds a custom 404.html that redirects folks back to `/index` with the path in the query string
- adds some logic to index.html that parses the parses the path out of the query string. 
- adds a caddy dev environment that should mimick github actions

I feel mixed about this PR. On the one hand, I think BrowserRouter produces prettier URLs and most importantly, those URLs support on-page anchor tags (needed for #31 ) But on the other hand, this PR is a bit of a hack and abuses the 404 page.

other solutions:
- static rendering (might help with SEO, e.g. right now google doesn't have any deep links into https://developers.nlx.ai/)
- from @peterszerzo on `npm run build`, copy `index.html` to all routes in `dist/`

### DIFF
--- a/packages/website/404.html
+++ b/packages/website/404.html
@@ -1,0 +1,39 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>NLX | Developers</title>
+    <script type="text/javascript">
+      // Single Page Apps for GitHub Pages
+      // MIT License
+      // https://github.com/rafgraph/spa-github-pages
+      // This script takes the current url and converts the path and query
+      // string into just a query string, and then redirects the browser
+      // to the new url with only a query string and hash fragment,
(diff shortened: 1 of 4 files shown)


peterszerzo (approved):

jakub-nlx:
It should be noted that this won't be a great solution for SEO, since GitHub will serve the 404 page with a 404 status and so crawlers will assume that its a dead link.

peterszerzo:
@jakub-nlx you're super right and @michaelglass and I are aware of the problem. The immediate goal was to make sure we can use hash links for things other than pages. Subsequently, we will server-render the pages.

PR #39, opened 2024-04-02
resolve eslints for cleaner documentation palette

michaelglass:
starting point for better docs

tsdoc / jsdoc is there to allow generated docs + more readable code. 

this is one step towards more readable code.

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -169,7 +169,7 @@ export interface ConversationHandler {
   sendWelcomeIntent: (context?: Context) => void;
   sendIntent: (intentId: string, context?: Context) => void;
   sendStructured: (request: StructuredRequest, context?: Context) => void;
-  subscribe: (subscriber: Subscriber) => void;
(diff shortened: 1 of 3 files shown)


packages/chat-core/src/index.ts
```diff
@@ -157,7 +157,7 @@ export interface ConversationHandler {
   sendWelcomeIntent: (context?: Context) => void;
   sendIntent: (intentId: string, context?: Context) => void;
   sendStructured: (request: StructuredRequest, context?: Context) => void;
-  subscribe: (subscriber: Subscriber) => void;
+  subscribe: (subscriber: Subscriber) => () => void;
```
peterszerzo, line 160: @michaelglass I don't understand how this is the correct type, the subscriber is already a function.

michaelglass: the subscribe function returns an unsubscribe function for that subscriber

peterszerzo: You're so right, thanks for straightening that out!

PR #45, opened 2024-04-04
Normalize slots dictionary as array

peterszerzo:
While the frontend should support both arrays and objects for slots in terms of API, the backend should always be sent arrays.

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -148,7 +148,7 @@ interface BotRequest {
     unstructured?: {
       text: string;
     };
-    structured?: StructuredRequest;
+    structured?: StructuredRequest & { slots?: SlotValue[] };
   };
 }
 
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -7,13 +7,14 @@ import { v4 as uuid } from "uuid";
 
 export type Context = Record<string, any>;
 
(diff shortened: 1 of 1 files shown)


michaelglass:
looks good. Have one amendment here 735627d477446bd9e1950216d5ba5c0a6b38ef5a. 

if you like it
`git fetch && git cherry-pick 735627d477446bd9e1950216d5ba5c0a6b38ef5a`

michaelglass (dismissed):

packages/chat-core/src/index.ts
```diff
@@ -148,7 +148,7 @@ interface BotRequest {
     unstructured?: {
       text: string;
     };
-    structured?: StructuredRequest;
+    structured?: StructuredRequest & { slots?: SlotValue[] };
```
jakub-nlx, line 151: Why is `slots` optional?

jakub-nlx (approved):

packages/chat-core/src/index.ts
michaelglass: I think this might be a question for the backend team. It looks like there's one endpoint which is responsible for a lot of different functionality.

peterszerzo: Because a structured request can combine `intentId`, `slots` or `choiceId` in the same requests, and all of these are optional (you can combine them every which way).

PR #46, opened 2024-04-05
Expiration check

peterszerzo:
If the conversation expired since the user interacted with the bot last, reset to a new ID.

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -1,6 +1,6 @@
 import fetch from "isomorphic-fetch";
 import ReconnectingWebSocket from "reconnecting-websocket";
-import { equals, adjust, omit } from "ramda";
+import { equals, adjust, omit, findLast } from "ramda";
 import { v4 as uuid } from "uuid";
 
 // Bot response
@@ -208,6 +208,20 @@ export const shouldReinitialize = (
   );
 };
 
+const getLastExpirationTimestamp = (
+  responses: Response[],
+): number | undefined => {
+  const lastExpirationResponse = findLast(
+    (response) =>
+      response.type === "bot" &&
+      typeof response.payload.expirationTimestamp === "number",
+    responses,
+  );
+  return lastExpirationResponse?.type === "bot"
+    ? lastExpirationResponse.payload.expirationTimestamp
+    : undefined;
+};
+
 export const createConversation = (config: Config): ConversationHandler => {
   let socket: ReconnectingWebSocket | undefined;
 
@@ -303,6 +317,15 @@ export const createConversation = (config: Config): ConversationHandler => {
     null;
 
   const sendToBot = async (body: BotRequest): Promise<void> => {
+    const lastExpirationTimestamp = getLastExpirationTimestamp(state.responses);
+    if (
+      typeof lastExpirationTimestamp === "number" &&
+      new Date().getTime() > lastExpirationTimestamp
+    ) {
+      setState({
+        conversationId: uuid(),
+      });
+    }
     const bodyWithContext = {
       userId: state.userId,
       conversationId: state.conversationId,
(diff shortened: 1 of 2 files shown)


peterszerzo:
Moved to drafts because https://github.com/nlxai/sdk/blob/7be3c14032b4ffcdd2f8a439a4e80d8f50a828ab/packages/chat-widget/src/index.tsx#L247-L279 is already handling this.

packages/chat-core/src/index.ts
```diff
@@ -221,6 +221,20 @@ export const shouldReinitialize = (
...
+): number | undefined => {
+  const lastExpirationResponse = findLast(
+    (response) =>
+      response.type === "bot" &&
+      typeof response.payload.expirationTimestamp === "number",
+    responses,
+  );
+  return lastExpirationResponse?.type === "bot"
+    ? lastExpirationResponse.payload.expirationTimestamp
+    : undefined;
```
michaelglass, line 235: minor nit (and less precise but)
```suggestion
return lastExpirationResponse?.payload?.expirationTimestamp;
```

packages/chat-core/src/index.ts
```diff
@@ -316,6 +330,15 @@ export const createConversation = (config: Config): ConversationHandler => {
     null;
 
   const sendToBot = async (body: BotRequest): Promise<void> => {
+    const lastExpirationTimestamp = getLastExpirationTimestamp(state.responses);
+    if (
+      typeof lastExpirationTimestamp === "number" &&
+      new Date().getTime() > lastExpirationTimestamp
+    ) {
+      setState({
+        conversationId: uuid(),
```
michaelglass, line 339: curious how this works for a multimodal when someone is already on the phone. Should this fail hard instead of resetting with a new conversation ID? What's the UX here

peterszerzo: This package does not cover multimodal.

michaelglass: oh yeah! whoops

packages/chat-core/src/index.ts
peterszerzo: I would rather do more precise and a bit clearer on the data structure.

packages/chat-core/src/index.ts
```diff
@@ -303,6 +317,15 @@ export const createConversation = (config: Config): ConversationHandler => {
     null;
 
   const sendToBot = async (body: BotRequest): Promise<void> => {
+    const lastExpirationTimestamp = getLastExpirationTimestamp(state.responses);
+    if (
+      typeof lastExpirationTimestamp === "number" &&
+      new Date().getTime() > lastExpirationTimestamp
+    ) {
+      setState({
```
michaelglass, line 325: 💭 this seems like it should raise an exception or signal in some way that there's a new conversation starting

michaelglass (approved):
the code here looks fine. I'm not sure if it's the right approach.

packages/chat-core/src/index.ts
peterszerzo: @michaelglass totally with you there, we are discussing the behind-the-scenes way to do this, basically any 'your conversation has expired' type of text content needs to be controlled inside the platform, so you can imagine some kind of explicit special message passing will happen here, but later.

sam-trost: Can we add a `conversationEnded` subscription thinger here.

sam-trost (approved):

packages/chat-core/src/index.ts
peterszerzo: @sam-trost that sounds like a bit of a change in direction, I would rather have a discussion about it at some point this week.

peterszerzo:
Abandoning this in favor of #124

PR #50, opened 2024-04-09
small api changes

michaelglass:
small code change while I'm writing documentation

feel free to review them commit-by-commit

1. simplify implementation of `shouldReinitialize` to make clear what fields are included / excluded from the logic.
2. ensure `nlx-api-key` is actually required

(I will keep adding small changes to this PR as I see them in the documentation PR / until this PR is merged)

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -103,7 +103,7 @@ export interface Config {
   responses?: Response[];
   failureMessage?: string;
   environment?: Environment;
-  headers?: Record<string, string>;
+  headers: Record<string, string> & { "nlx-api-key": string };
   languageCode: string;
   // Experimental settings
   experimental?: {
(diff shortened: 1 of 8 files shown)


peterszerzo (approved):

packages/chat-core/src/index.ts
```diff
@@ -202,22 +202,9 @@ export const shouldReinitialize = (
...
-    ) ||
-    !equals(
-      config1.experimental?.completeBotUrl,
-      config2.experimental?.completeBotUrl,
-    ) ||
-    !equals(config1.headers, config2.headers) ||
-    !equals(config1.responses, config2.responses) ||
-    !equals(config1.environment, config2.environment)
+  return !equals(
+    omit(["failureMessage"], config1),
```
michaelglass, line 206: @peterszerzo actually, based on usage, I guess this should reinitialize if config changes in any way, so `!equals(config1, config2)` should be correct.

peterszerzo: @michaelglass you're right, let's have it do that, feel free to open a PR :)

PR #55, opened 2024-04-17
document core chat api

michaelglass:
flesh out tsdoc for chat-core library

I'm unhappy with the formatting & ordering of the auto-generated docs.

Some options:
- stop using markdown rendering, and render typedoc natively to html on a different domain, e.g. `api.developers.nlx.ai`
- use typedoc to render to json, and write our own typedoc renderer to handle formatting & ordering ourself
- manually massage the markdown output until it looks more acceptable.

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -301,7 +301,7 @@ export interface Config {
    */
   conversationId?: string;
   /**
-   * TODO I don't know what this is
+   * Setting the `userID` allows it to be searchable in bot history, as well as usable via `{System.userId}` in the intent.
    */
   userId?: string;
   /**
@@ -954,7 +954,6 @@ export function promisify<T>(
   return async (payload: T) => {
     return await new Promise((resolve, reject) => {
       const timeoutId = setTimeout(() => {
-        // TODO should this unsubscribe?
         reject(new Error("The request timed out."));
       }, timeout);
       const subscription = (
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -4,510 +4,15 @@ import { equals, adjust, omit } from "ramda";
 import { v4 as uuid } from "uuid";
 import packageJson from "../package.json";
 
-// Bot response
-
 /**
- * Context to send back to the bot, for usage later in the intent.
+ * Call this to create a conversation handler.
+ *
+ * @param config -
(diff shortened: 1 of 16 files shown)


packages/chat-core/src/index.ts
```diff
@@ -4,113 +4,343 @@ import { equals, adjust, omit } from "ramda";
...
+  /**
+   * Fetch this from the bot's Deployment page.
+   */
   botUrl: string;
+  /**
+   * The conversation ID. If not set, a new conversation will be started.
+   */
   conversationId?: string;
+  /**
+   * TODO I don't know what this is
```
michaelglass, line 304: AFAICT this is to help user's find their own user ids? E.g. this is not a NLX User id but rather a an arbitrary, optional user id in a customer's system.

packages/chat-core/src/index.ts
```diff
@@ -586,6 +954,7 @@ export function promisify<T>(
   return async (payload: T) => {
     return await new Promise((resolve, reject) => {
       const timeoutId = setTimeout(() => {
+        // TODO should this unsubscribe?
```
michaelglass, line 957: should this timeout also call `convo.unsubscribe(subscription);` ?

peterszerzo: Yes it should!

packages/chat-core/src/index.ts
peterszerzo: Basically, if you give us a user ID, the conversation logs will be searchable by it if the privacy agreement allows. It will also be usable inside the canvas (`{System.userId}` inside messages, payloads etc.).

peterszerzo (approved):

peterszerzo (approved):

peterszerzo (approved):

packages/chat-core/src/index.ts
```diff
@@ -119,8 +350,6 @@ const welcomeIntent = "NLX.Welcome";
 
 const defaultFailureMessage = "We encountered an issue. Please try again soon.";
 
-export type State = Response[];
```
michaelglass, line 122: API change

packages/chat-core/src/index.ts
```diff
@@ -196,20 +525,35 @@ const safeJsonParse = (val: string): any => {
   }
 };
 
-type Subscriber = (response: Response[], newResponse?: Response) => void;
```
michaelglass, line 199: this becomes exported

PR #56, opened 2024-04-17
small sdk tweaks

michaelglass:
two small follow-ups from #55, #50

- fix  a bug in `promisify`
- change reinitialize logic

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -584,6 +584,7 @@ export function promisify<T>(
     return await new Promise((resolve, reject) => {
       const timeoutId = setTimeout(() => {
         reject(new Error("The request timed out."));
+        convo.unsubscribe(subscription);
       }, timeout);
       const subscription = (
         _responses: Response[],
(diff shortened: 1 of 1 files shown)


packages/chat-core/src/index.ts
```diff
@@ -587,6 +584,7 @@ export function promisify<T>(
     return await new Promise((resolve, reject) => {
       const timeoutId = setTimeout(() => {
         reject(new Error("The request timed out."));
+        convo.unsubscribe(subscription);
```
peterszerzo, line 587: It's ok if `subscription` is defined below right? On account of the callback?

michaelglass: shouldl be fine 

test code

```
const a = () => {
    console.log(b);
}
const b = 'foo';
a();
```

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDAFASiQPhgbwFAyEyiQgA2ApgHRkgDmKARmgNx4C+eJ0MjSMAcgBmIEALZx0LIA

peterszerzo (approved):

PR #61, opened 2024-05-01
Add disabled input feature

peterszerzo:
Disable the input field if the last payload contains `nlx:input-disabled=true`

Screenshot 2024-05-01 at 10 09 59 AM

### DIFF
--- a/packages/chat-widget/src/index.tsx
+++ b/packages/chat-widget/src/index.tsx
@@ -24,6 +24,7 @@ import {
   SendIcon,
   ErrorOutlineIcon,
 } from "./icons";
+import { last } from "ramda";
 import * as constants from "./ui/constants";
 import {
   type Props,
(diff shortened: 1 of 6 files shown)


packages/chat-widget/src/index.tsx
```diff
@@ -325,6 +326,20 @@ const ConversationHandlerContext = createContext<ConversationHandler | null>(
   null,
 );
 
+// eslint-disable-next-line jsdoc/require-returns
```
michaelglass, line 329: this eslint-disable-next-line is unecessary if the const isn't exported

michaelglass (approved):
looks good!

packages/chat-widget/src/index.tsx
peterszerzo: @michaelglass it wasn't passing...

PR #60, opened 2024-04-29
Journey builder

peterszerzo:
Alternative to #32, on a new branch in order to preserve the previous version after a tricky rebase.

PR #68, opened 2024-05-06
Journey manager package

peterszerzo:
Alternative to #60, with linting and tsdoc.

### DIFF
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -1,0 +1,118 @@
+import { type VoiceCompass } from "@nlxai/voice-compass";
+import { find, decode, Query, EncodedQuery } from "./queries";
+
+type StepId = string;
+
+// TODO: add regex match condition
+type UrlCondition = {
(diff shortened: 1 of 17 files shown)


packages/journey-manager/src/index.ts
```diff
@@ -0,0 +1,166 @@
...
+  event: "pageLoad" | "click";
+  /**
+   * A query identifying the element
+   */
+  query?: EncodedQuery;
+  /**
+   * URL condition
+   */
+  urlCondition?: UrlCondition;
+}
```
jakub-nlx, line 40: Probably nicer as a discriminated union as "pageLoad" doesn't have a query and "click" always has a query:

```suggestion
export type Trigger = {
  /**
   * Event
   */
  event: "pageLoad";
  /**
   * URL condition
   */
  urlCondition?: UrlCondition;
} | {
/**
   * Event
   */
  event: "click";
  /**
   * A query identifying the element
   */
  query: EncodedQuery;
  /**
   * URL condition
   */
  urlCondition?: UrlCondition;
}
```

jakub-nlx (approved):

PR #70, opened 2024-05-07
monorepo tsconfig

michaelglass:
- **disable strictNullChecks because strict is enough**
- **enable forceConsistentCasingInFileNames**
- **chat-core uses shared tsconfig**
- **consume shared tsconfig everywhere**
- **remove options in website**
- **use modern react jsx transpilation**
- **drop unnecessary lib in chat-widget tsconfig**


packages/chat-widget/src/ui/components.tsx
```diff
@@ -475,6 +475,12 @@ export const PinBubble: React.FunctionComponent<{
   </PinBubbleContainer>
 );
 
+PinBubble.propTypes = {
```
peterszerzo, line 478: @michaelglass what is the rationale behind prop types?

michaelglass: added a less intrusive fix for what was a lint failure.

PR #72, opened 2024-05-14
Add once modifier. Add CSS selector

peterszerzo:


packages/journey-manager/src/index.ts
```diff
@@ -87,6 +116,22 @@ const matchesUrlCondition = (urlCondition: UrlCondition): boolean => {
...
+  const triggeredSteps = getTriggeredSteps(config.conversationId);
+
+  const sendStep = (stepId: string, once: boolean): void => {
+    if (triggeredSteps.includes(stepId)) {
+      triggeredSteps.push(stepId);
+      saveTriggeredSteps(config.conversationId, triggeredSteps);
+      if (once) {
+        return;
+      }
+    }
```
jakub-nlx, line 128: Something doesn't seem quite right here. How would a stepId ever get into `triggeredSteps` in the first place? Perhaps a `!` missing in the conditional?

packages/journey-manager/src/queries.ts
```diff
@@ -126,12 +126,27 @@ export async function evaluate(q: Query): Promise<boolean> {
...
 
-  const methodName: keyof typeof queryFns = `findBy${q.queryName}`;
-
-  return await queryFns[methodName](container, ...(q.queryArgs as QueryArgs));
+  let results: HTMLElement[][] = [];
+  if (q.queryName === "QuerySelector") {
+    results = await Promise.all(
+      containers.map((container) => [
+        ...container.querySelectorAll<HTMLElement>(q.queryArgs[0] as string),
+      ]),
```
jakub-nlx, line 137:
```suggestion
       containers.map((container) =>
        waitFor(
          () => {
            const els = [
              ...container.querySelectorAll(q.queryArgs[0] as string),
            ];
            if (els.length == 0) {
              throw new Error("No elements found");
            }
          },
          { container },
        ),
      ),
```

where `waitFor` comes from `"@testing-library/dom"`

jakub-nlx (changes requested):

packages/journey-manager/src/queries.ts
peterszerzo: @jakub-nlx how do you get away without returning `els`, am I missing something? This is what I could come up with:

```ts
      containers.map((container) =>
        waitFor(
          () => {
            const els = [
              ...container.querySelectorAll<HTMLElement>(
                q.queryArgs[0] as string,
              ),
            ];
            if (els.length == 0) {
              throw new Error("No elements found");
            }
            return els;
          },
          { container },
        ),

```

packages/journey-manager/src/index.ts
peterszerzo: You're totally right, updated the logic!

packages/journey-manager/src/queries.ts
jakub-nlx: Oh yeah, you definitely want to return `els` 🤦

packages/journey-manager/src/queries.ts
```diff
@@ -126,12 +127,42 @@ export async function evaluate(q: Query): Promise<boolean> {
...
+            { container },
+          ),
+      ),
+    );
+  } else {
+    const methodName: keyof typeof queryFns = `findAllBy${q.queryName}`;
+    results = await Promise.all(
+      containers.map(
+        async (container) =>
+          // TODO improve typing
```
jakub-nlx, line 160:
```suggestion
```

Typing now seems fine, no?

jakub-nlx (approved):

jakub-nlx (approved):

PR #74, opened 2024-05-15
Builder api extension

peterszerzo:
- expand on `journeyManager.run` output.
- improve typedocs.
- set up documentation chapter.


packages/journey-manager/src/queries.ts
```diff
@@ -18,11 +18,25 @@ export interface Query {
    */
   queryArgs: [string | RegExp, Record<string, RegExp | boolean>?];
   /**
-   *
+   * Parent query
```
jakub-nlx, line 21:
```suggestion
   * If present, contains the query to only the inside of an element found by the parent query.
```

jakub-nlx (approved):
We should have some documentation somewhere on how the query engine actually works. Not sure that's developer documentation or user documentation though...

PR #75, opened 2024-05-15
Digression callback

peterszerzo:
Naive implementation of digression detection.

### DIFF
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -127,15 +127,50 @@ export interface RunOutput {
 }
 
 /**
+ * Configuration for the run method
+ */
+export interface RunProps {
+  /**
+   * The regular multimodal configuration
+   */
+  config: Config;
+  /**
+   * The triggers dictionary, downloaded from the Dialog Studio desktop app
+   */
+  triggers: Triggers;
+  /**
+   * Digression detection callback
+   */
+  onDigression?: (client: Client) => void;
+}
+
+/**
  * Run the multimodal journey
- * @param config - The voice compass configuration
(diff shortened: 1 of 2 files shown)


packages/journey-manager/src/index.ts
```diff
@@ -126,17 +126,48 @@ export interface RunOutput {
...
  * @param config - The voice compass configuration
  * @param triggers - The triggers dictionary, downloaded from the Dialog Studio desktop app
+ * @param options - Journey run configuration
  * @returns an object containing a teardown function and the multimodal client.
  */
-export const run = (config: Config, triggers: Triggers): RunOutput => {
+export const run = (
+  config: Config,
+  triggers: Triggers,
+  options?: Options,
```
jakub-nlx, line 149: Kind of weird having both `Config` and `Options`, they both sound kind of like bags of stuff. Maybe it would be nicer if we merged them?

peterszerzo: The `Config` is the same config that the regular voice compass SDK is using, we could do some kind of `Config & Options` thing but I think it's more awkward.

jakub-nlx (dismissed):

packages/journey-manager/src/index.ts
jakub-nlx: Maybe then using a slightly better name like `ConnectionParameters` or some such

peterszerzo:
@jakub-nlx I reworked things a bit to adhere more to the way the widget is initialized (there we have this same thing that there is a configuration that is identical to that of lower level apis): https://developers.nlx.ai/widget-setup

jakub-nlx (approved):

PR #77, opened 2024-05-16
Adds some conceptual documentation to the query engine

jakub-nlx:
I don't know if this is the most suitable place, but I feel like users should have access to this in one place or another (or at least our own staff).


peterszerzo (approved):

peterszerzo:
@jakub-nlx great stuff!!!

PR #86, opened 2024-05-30
Add nudge configuration options

peterszerzo:


packages/chat-widget/src/index.tsx
```diff
@@ -495,22 +495,25 @@ export const Widget = forwardRef<WidgetRef, Props>(function Widget(props, ref) {
 
   // Bubble
 
-  const [bubble, setBubble] = useState(false);
+  const [isNudgeVisible, setIsNudgeVisible] = useState(false);
+
+  const showNudgeAfter = props.nudge?.showAfter ?? 3_000;
+  const hideNudgeAfter = props.nudge?.hideAfter ?? 20_000;
```
jakub-nlx, line 501:
```suggestion
  const hideNudgeAfter = Math.max(showNudgeAfter, props.nudge?.hideAfter ?? 20_000);
```

or perhaps it would be better to just change the useEffect to:

```
useEffect(() => {
  let timeout = setTimeout(() => {
     setIsNudgeVisible(true);
     timeout = setTimeout(() => {
        setIsNudgeVisible(false);
     });
  }, showNudgeAfter);

 return () => {
   clearTimeout(timeout);
 };
}, [showNudgeAfter, hideNudgeAfter]);
```

That way the `hideNudgeAfter` is always guaranteed to be called after the showNudgeAfter, but also the amount of time is perhaps more logically after showing rather than from the origin point...

peterszerzo: @jakub-nlx I'm using your code snippet below, just had to add the `hideNudgeAfter` argument to the second timeout.

jakub-nlx (dismissed):

packages/chat-widget/src/index.tsx
jakub-nlx: Oh yeah.

jakub-nlx (approved):

PR #87, opened 2024-05-31
Support markdown in nudge. Set up nudge close callback

peterszerzo:


packages/chat-widget/src/index.tsx
```diff
@@ -554,6 +554,7 @@ export const Widget = forwardRef<WidgetRef, Props>(function Widget(props, ref) {
             <C.PinBubble
               isActive={!expanded && isNudgeVisible}
               onClick={() => {
+                props.onNudgeClose?.(chat.conversationHandler);
```
jakub-nlx, line 557: Should this not also be called when the timeout triggers?

peterszerzo: I don't think so, they want to know how many times the user dismisses it explicitly, for analytics purposes. We might need separate events but I would rather kick that can down the road.

jakub-nlx (changes requested):
Also the docs for `Nudge.content` should point out that they support markdown.

(And perhaps point out that there is a security risk in uncontrolled strings being passed in there).


Also (and this is more about this feature in general), wouldn't it be more flexible to just accept arbitrary jsx as the content? That way the consumer of this code can use Markdown if they wish, but can also easily produce other kinds of content (like svg images as react components) without having to serialize them to strings first...

packages/chat-widget/src/props.ts
```diff
@@ -139,6 +139,10 @@ export interface Props {
    * Optional callback to be called when the chat is closed via the close button.
    */
   onClose?: () => void;
+  /**
+   * Optional callback to be called when the nudge element is closed.
+   */
+  onNudgeClose?: (conversationHandler: ConversationHandler) => void;
```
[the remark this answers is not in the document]
peterszerzo: I'm not sure at this point, all of these event handlers are all top-level and I'm also thinking at some point we could just do a `widget.addEventListener()` or `widget.on()` to not bloat this configuration object too much (`widget` being the object returned by the call with the giant configuration object).

peterszerzo:
I would not encourage jsx because that would mean they can (and boy they will) add all kinds of styles that mess up the looks/layout of the nudge. They do it anyway in markdown but there it's easier to make the case with examples that say 'just add regular text stuff, maybe a newline character, maybe a link'.

peterszerzo:
@jakub-nlx added a note with markdown support to the `nudge.content` docs.

packages/chat-widget/src/props.ts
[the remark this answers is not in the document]
jakub-nlx: Well it seems a bit strange to have everything related to the nudge in the nudge object, but the callback separate. It also leads you to strange territory from a type perspective, for instance you can now have `onNudgeClose` set, but no `nudge` present, which is sort of bizarre (albeit probably not harmful in any way in this instance).

packages/chat-widget/src/index.tsx
jakub-nlx: Maybe that should be then documented. From the current docs, I wouldn't understand that it's just for manual closes.

peterszerzo: Pushes a commit for the docs.

packages/chat-widget/src/props.ts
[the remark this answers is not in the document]
peterszerzo: It's a good point - I would make a mental note of this and have a discussion on what we do with events in general. We have some wiggle room before a 1.0 release.

PR #90, opened 2024-06-04
Hard-codes an input color

jakub-nlx:
This seems to be a more reliable fix in practice


michaelglass (approved):
looks good. I think forcing color scheme might also make sense

peterszerzo (approved):

PR #91, opened 2024-06-10
Find active triggers on page

peterszerzo:


packages/journey-manager/src/index.ts
```diff
@@ -120,6 +130,10 @@ export interface RunOutput {
    * Stop running the journey, removing all event listeners
    */
   teardown: () => void;
+  /**
+   * Find active triggers on the page
+   */
+  findActiveTriggers: () => Promise<ActiveTriggers>;
```
jakub-nlx, line 136: I'm sort of struggling to imagine exactly how this will be used in practice. The promise in the interface makes this more puzzling, since it's not particularly clear why this should be about some future state. 

Wouldn't it be a lot easier to use this sort of thing if we gave the users a callback whenever an element became an active trigger and another when it stopped being one?

peterszerzo: The promise is just because `find` is async, implementation-wise it does not seem possible. The callback does sound nice but I would imagine it would come with some management overhead on the users' side, maintaining a full list at all times (I guess we could do this for them). Then again, not sure if we need to go down the whole mutation observer thing and re-run the find method if the website could just find relevant ID's whenever it needs to (if a digression happened, if inactivity happened etc.).

jakub-nlx: Well, I'd suggest implementing an example of a typical use case before publishing this, as I at least can't quite imagine how I'd build those features using this primitive without essentially having to build element-wise diffing.

Regarding asynchrony: we have the `getAll` function that is sync (although I notice that doesn't for some reason support the `parent` parameter, although that's a fairly trivial omission to fix).

peterszerzo: `findActiveTriggers().click.forEach(node => node.classList.add(""))` is all it takes to do the simplest use case on this feature. Then if you want to run it again later, you store the modified nodes, remove the extra stuff from them, then add to the ones from the fresh call. Some might get added and removed but I doubt it would happen very much.

As I'm writing this out, I'm realizing that you do have a point on getting callbacks for activate and deactivate in terms of avoiding stale stuff and avoiding the need for the whole memorizing trick, I'm just not sure it's feasible on timing. Maybe we could set up the API that way but not quite yet implement the mutation observer?

Let's talk tomorrow to make a call!

jakub-nlx (approved):

PR #94, opened 2024-06-12
Set up journey manager UI

peterszerzo:
Implements https://linear.app/nlxai/issue/FRO-7/journey-manager-ui-prototype

### DIFF
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -218,6 +218,7 @@ export const run = (props: RunProps): RunOutput => {
 
   if (props.ui != null) {
     uiElement = document.createElement("journey-manager");
+    uiElement.style.zIndex = 1000;
     uiElement.config = props.ui;
     uiElement.client = client;
     document.body.appendChild(uiElement);
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -218,7 +218,6 @@ export const run = (props: RunProps): RunOutput => {
 
   if (props.ui != null) {
     uiElement = document.createElement("journey-manager");
-    console.log(uiElement, props.ui);
     uiElement.config = props.ui;
     uiElement.client = client;
     document.body.appendChild(uiElement);
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -1,10 +1,20 @@
 import { type Config, type Client, create } from "@nlxai/voice-compass";
 import { find, getAll, decode, type Query, type EncodedQuery } from "./queries";
+import { type UiConfig, JourneyManagerElement } from "./ui";
 export {
+  type Query,
   type EncodedQuery,
   type Method,
   type SerializedRegex,
 } from "./queries";
+export {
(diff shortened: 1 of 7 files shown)


packages/journey-manager/src/index.ts
```diff
@@ -183,6 +213,17 @@ function filterMap<X, Y>(
 export const run = (props: RunProps): RunOutput => {
   const client = create(props.config);
 
+  // TODO: type this more accurately
+  let uiElement: any;
+
+  if (props.ui != null) {
+    uiElement = document.createElement("journey-manager");
+    console.log(uiElement, props.ui);
```
jakub-nlx, line 221:
```suggestion
```

packages/journey-manager/src/ui.ts
```diff
@@ -0,0 +1,380 @@
...
+  /**
+   * Font family
+   */
+  fontFamily: string;
+}
+
+/**
+ * Deep partial variant of the UI theme, input by the library user
+ */
+export interface PartialTheme {
```
jakub-nlx, line 35:
```typescript
export type DeepPartial<T> = {
  [P in keyof T]?: T[P] extends Array<infer I>
    ? Array<DeepPartial<I>>
    : DeepPartial<T[P]>;
};
```

Then you can just say `DeepPartial<Theme>`

packages/journey-manager/src/ui.ts
```diff
@@ -0,0 +1,380 @@
...
+const mergeWithDefault = (partial?: PartialTheme): Theme => ({
+  ...defaultTheme,
+  ...(partial ?? {}),
+  colors: {
+    ...defaultTheme.colors,
+    ...(partial?.colors ?? {}),
+  },
+});
+
+const styles = (theme: Theme): string => `
```
jakub-nlx, line 89: I'm kind of curious why use javascript for the theme-ing rather than CSS custom properties?

packages/journey-manager/src/ui.ts
```diff
@@ -0,0 +1,380 @@
...
+      }
+    };
+
+    pinButton?.addEventListener("click", handlePinButtonClick);
+    drawer?.addEventListener("click", handleDrawerClick);
+
+    this._teardown = () => {
+      pinButton?.removeEventListener("click", handlePinButtonClick);
+      drawer?.removeEventListener("click", handleDrawerClick);
+    };
```
jakub-nlx, line 371: Is this necessary? Presumably being removed from the DOM cleans up event listeners?

peterszerzo: I don't think so, how sure would you say you are? I realize this is not too important anyway, just doing the discipline thing :).

packages/journey-manager/src/index.ts
peterszerzo: Removed in a follow-up commit.

packages/journey-manager/src/ui.ts
peterszerzo: Offloading this to https://github.com/nlxai/sdk/issues/95.

jakub-nlx (approved):

PR #97, opened 2024-06-12
Fix button event handling

peterszerzo:


jakub-nlx (approved):
Yeah I think this will do the trick as well.

PR #99, opened 2024-06-14
Rename voice-compass to multimodal

peterszerzo:
Fixes https://linear.app/nlxai/issue/FRO-14/rename-voice-compass-package-to-multimodal.


jakub-nlx (approved):
We should also [deprecate the old version](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions).

peterszerzo:
@jakub-nlx absolutely!!

PR #100, opened 2024-06-17
Fix promisify wrapper

peterszerzo:
Resolve chat request promises on failure (not just on proper bot response)

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -961,7 +961,7 @@ export function promisify<T>(
         _responses: Response[],
         newResponse: Response | undefined,
       ): void => {
-        if (newResponse?.type === "bot") {
+        if (newResponse?.type === "bot" || newResponse?.type === "failure") {
           clearTimeout(timeoutId);
           convo.unsubscribe(subscription);
           resolve(newResponse);


packages/chat-core/src/index.ts
```diff
@@ -961,7 +961,7 @@ export function promisify<T>(
         _responses: Response[],
         newResponse: Response | undefined,
       ): void => {
-        if (newResponse?.type === "bot") {
+        if (newResponse?.type === "bot" || newResponse?.type === "failure") {
           clearTimeout(timeoutId);
           convo.unsubscribe(subscription);
           resolve(newResponse);
```
michaelglass, line 967: should it resolve or fail here?

michaelglass (approved):

packages/chat-core/src/index.ts
peterszerzo: I don't think the error can be typed accurately, so my take is that unless we can do something about that, definitely no rejection.

michaelglass: yeah, trade-offs. 

On the one hand: better typechecking through the whole system, but two ways to have errors

On the other hand: a single way to detect errors that's enforced through the language, but forcing type-checking to, e.g. check if the error object has a `response` field or something.

peterszerzo: I'm not exactly married to my opinion above especially since failure could just reject with a string (that's pretty much what it is), the road I don't feel good about walking down on is more like the philosophical rabbit hole of defining what an error is. There are cases where bot-type responses technically mean failure (it is in the message metadata), which would end up in the `.then` branch at the moment, and it seems very cumbersome to separate it out of there.

peterszerzo: I will merge this PR now, @michaelglass and @jakub-nlx feel free to suggest alternative API's at any time with this discussion in mind.

PR #101, opened 2024-06-17
Adding trigger "once" value into loadSteps

will-nlx:
## Summary of Issues
```
loadSteps.forEach(({ stepId, urlCondition, once }) => {
    if (urlCondition != null && !matchesUrlCondition(urlCondition)) {
      return;
    }
    sendStep(stepId, once ?? false);
  });
  ```

`once ?? false` always resolving to false because trigger's `once` value isn't being captured. 

## Fix
Capture `trigger.once` value during `loadSteps` array creation


peterszerzo (approved):

jakub-nlx:
Nice!

As a separate idea, perhaps an improvement would be to make the `once` not optional in the `ClickStep` or `LoadStep` types. It's optional earlier on as a convenience, but there's no reason for `once` to take 3 values rather than 2 once we're processing it.

peterszerzo:
@jakub-nlx totally, that's much better!

PR #102, opened 2024-06-18
Add trigger highlights

peterszerzo:
Fixes https://linear.app/nlxai/issue/FRO-17/pulsing-highlights-on-journey-manager.

@jakub-nlx this removes `findActiveTriggers` from the public API as it becomes internal to the package. So we can improve on the implementation later without having the end user have to change anything. Lame 500ms setInterval-based implementation for now, but only triggers if highlights are enabled.

https://github.com/nlxai/sdk/assets/6738398/d40a1677-ba1c-47cb-933f-88b256eb13e5

### DIFF
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -161,10 +161,6 @@ export interface RunOutput {
    */
   teardown: () => void;
   /**
-   * Find active triggers on the page
-   */
-  findActiveTriggers: (eventType: ActiveTriggerEventType) => ActiveTrigger[];
-  /**
    * The regular multimodal SDK client
    */
   client: Client;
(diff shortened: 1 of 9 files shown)


packages/journey-manager/src/index.ts
```diff
@@ -376,12 +326,73 @@ export const run = (props: RunProps): RunOutput => {
     return [];
   };
 
+  /**
+   * UI management
+   */
+
+  // TODO: type this more accurately
+  let uiElement: any;
```
jakub-nlx, line 334:
```suggestion
  let uiElement: JourneyManagerElement | undefined;
```

packages/journey-manager/src/index.ts
```diff
@@ -376,12 +326,73 @@ export const run = (props: RunProps): RunOutput => {
...
+    };
+    let highlightInterval: NodeJS.Timeout | null = null;
+    if (props.ui.highlights === true) {
+      highlightInterval = setInterval(() => {
+        const highlightElements = findActiveTriggers("click").flatMap(
+          (activeTrigger) => activeTrigger.elements,
+        );
+        uiElement.highlightElements = highlightElements;
+      }, 500);
+    }
```
jakub-nlx, line 380: Why not mutation observer? In fact, we could probably even build a faster `matchesActiveTriggers` that would fit neatly into mutation observer callbacks.

packages/journey-manager/src/index.ts
peterszerzo: I couldn't make it work... what the `any` helps with is that it foregoes having to type custom events.

packages/journey-manager/src/index.ts
```diff
@@ -160,10 +160,6 @@ export interface RunOutput {
    * Stop running the journey, removing all event listeners
    */
   teardown: () => void;
-  /**
-   * Find active triggers on the page
-   */
-  findActiveTriggers: (eventType: ActiveTriggerEventType) => ActiveTrigger[];
```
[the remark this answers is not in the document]
peterszerzo: Isn't this an API that you didn't feel good about, preferring a subscription model instead? Just curious how you imagine the entire API surface of this package looking like.

From my end this was a pragmatic step.

packages/journey-manager/src/index.ts
jakub-nlx: Ah ok.

jakub-nlx (approved):

packages/journey-manager/src/index.ts
peterszerzo: Here it is: https://github.com/nlxai/sdk/issues/103

PR #104, opened 2024-06-19
Fix triggers retrieve method

peterszerzo:
Seems like a logical mistake.


peterszerzo:
@jakub-nlx merging this, feel free to review after-the-fact.

gampleman (commented):
Yeah LGTM

PR #105, opened 2024-06-19
Redirect if the previous step happened on a different URL

peterszerzo:
Fixes https://linear.app/nlxai/issue/FRO-30/handle-previous-step-button-in-control-center-ui.

### DIFF
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -117,11 +117,19 @@ const matchesUrlCondition = (urlCondition: UrlCondition): boolean => {
 const localStorageKey = (conversationId: string): string =>
   `jb-triggered-steps-${conversationId}`;
 
-const saveTriggeredSteps = (conversationId: string, steps: string[]): void => {
+const saveTriggeredSteps = (
+  conversationId: string,
+  steps: TriggeredStep[],
+): void => {
   localStorage.setItem(localStorageKey(conversationId), JSON.stringify(steps));
 };
 
-const getTriggeredSteps = (conversationId: string): string[] => {
+interface TriggeredStep {
+  stepId: string;
+  url: string;
+}
+
+const getTriggeredSteps = (conversationId: string): TriggeredStep[] => {
   try {
     const jsonString = localStorage.getItem(localStorageKey(conversationId));
     if (jsonString == null) {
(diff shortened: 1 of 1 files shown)


packages/journey-manager/src/index.ts
```diff
@@ -363,10 +373,14 @@ export const run = (props: RunProps): RunOutput => {
       if (action === "previous") {
         const lastTriggeredStep = triggeredSteps[triggeredSteps.length - 1];
         if (lastTriggeredStep != null) {
-          client.sendStep(lastTriggeredStep).catch((err) => {
+          client.sendStep(lastTriggeredStep.stepId).catch((err) => {
             // eslint-disable-next-line no-console
             console.warn(err);
           });
+          // Redirect to previous page if the last triggered step occurred on it
+          if (lastTriggeredStep.url !== window.location.toString()) {
```
michaelglass, line 381: looks good. just double checked: changing anchor tags don't trigger a reload.

michaelglass (approved):

packages/journey-manager/src/index.ts
```diff
@@ -226,12 +234,14 @@ export const run = (props: RunProps): RunOutput => {
   }
 
   const sendStep = (stepId: string, once: boolean): void => {
-    if (triggeredSteps.includes(stepId)) {
+    if (
+      triggeredSteps.some((triggeredStep) => triggeredStep.stepId === stepId)
+    ) {
       if (once) {
```
jakub-nlx, line 240: _perf:_ probably would be more performant to do:

```javascript
if (once && triggeredSteps.some((triggeredStep) => triggeredStep.stepId === stepId))
```

(i.e. avoid checking the array if we didn't call it with `once`.

packages/journey-manager/src/index.ts
```diff
@@ -363,10 +373,14 @@ export const run = (props: RunProps): RunOutput => {
...
         const lastTriggeredStep = triggeredSteps[triggeredSteps.length - 1];
         if (lastTriggeredStep != null) {
-          client.sendStep(lastTriggeredStep).catch((err) => {
+          client.sendStep(lastTriggeredStep.stepId).catch((err) => {
             // eslint-disable-next-line no-console
             console.warn(err);
           });
+          // Redirect to previous page if the last triggered step occurred on it
+          if (lastTriggeredStep.url !== window.location.toString()) {
+            window.location.href = lastTriggeredStep.url;
```
jakub-nlx, line 382: Hm... 

Should we check that the trigger isn't present on the existing page (assuming the existing page also matches that triggers url filter?)

I feel like there could be a bunch of scenarios where the URL is updated with various state query parameters etc, which might cause trouble...

PR #106, opened 2024-06-20
Support digressions when steps without url conditions are used for escalate/end

peterszerzo:


packages/journey-manager/src/index.ts
```diff
@@ -211,17 +211,29 @@ export const run = (props: RunProps): RunOutput => {
 
   const triggeredSteps = getTriggeredSteps(props.config.conversationId);
 
-  // If all triggers have a URL constraint set and none match the current URL, fire a digression calback
+  /**
+   * Digression detection
+   * - if all triggers have a URL constraint set and none match the current URL, fire a digression callback
+   * - notable exception: if the step is used as escalation or end as well, missing URL constraint is ok.
```
michaelglass, line 217: rewriting to confirm I understand what you're saying
```suggestion
   * - notable exception: ignore URL constraints if if the step is used as an escalation or end.
```

packages/journey-manager/src/index.ts
```diff
@@ -211,17 +211,29 @@ export const run = (props: RunProps): RunOutput => {
...
+  const isDigressionDetectable = !Object.entries(props.triggers).some(
+    ([stepId, trigger]) => {
+      return (
+        trigger.urlCondition == null &&
+        !(stepId === props.ui?.escalationStep || stepId === props.ui?.endStep)
+      );
+    },
+  );
+
+  if (isDigressionDetectable && !urlConditions.some(matchesUrlCondition)) {
```
michaelglass, line 236: 💭 minor nit: you can avoid all of the computation by checking if `onDigression` is set before checking for a digression.

packages/journey-manager/src/index.ts
```diff
@@ -211,17 +211,29 @@ export const run = (props: RunProps): RunOutput => {
...
     (trigger) => trigger.urlCondition,
   );
 
-  if (
-    Object.keys(props.triggers).length === urlConditions.length &&
-    !urlConditions.some(matchesUrlCondition)
-  ) {
+  // If there are any steps for which there is no URL condition while also not being used for escalation or end,
+  // the package assumes that a digression cannot be reliably detected.
+  const isDigressionDetectable = !Object.entries(props.triggers).some(
```
michaelglass, line 227: suggest rewriting this as an `every` for readability

```suggestion
  const isDigressionDetectable = !Object.entries(props.triggers).every(
    ([stepId, trigger]) => {
      return (
        // every step has a URL condition
        trigger.urlCondition != null ||
        // unless it's an escalation step or end step
        stepId === props.ui?.escalationStep ||
        stepId === props.ui?.endStep
      );
    },
  );
```

michaelglass (dismissed):
some stylistic suggestions. Take them or leave them.

packages/journey-manager/src/index.ts
peterszerzo: In the future there will be automatic digression handling (later today actually), I think I would leave this as it is.

packages/journey-manager/src/index.ts
peterszerzo: @michaelglass you meant to take out the `!` from the `Object.entries` here right?

michaelglass: whooops, yeah!

jakub-nlx (dismissed):

packages/journey-manager/src/index.ts
peterszerzo: I agree, so nice @michaelglass 😍

jakub-nlx (approved):

PR #107, opened 2024-06-20
Position journey manager custom element as absolute

peterszerzo:
Helps Safari compatibility.


michaelglass (approved):
🤞 assume this works. I'm not sure how to test it myself but it's certainly not a security breach so this review complies with SOC2 🤷

PR #110, opened 2024-06-21
Recompute journey manager highlights on element change

peterszerzo:
If the element changes, positions need to be recomputed.


jakub-nlx (approved):
What if the element changes but its identity stays the same? Like an animation or something? I think floating-ui has a subscription feature for that case, but that would definitely complicate matters.

peterszerzo:
@jakub-nlx I think we will have to do both and that's a can of worms for next week...

Issue #93, opened 2024-06-11
Move 'DOMContentLoaded' handling inside journey manager


peterszerzo:
In order to avoid relying on client code running journey manager on content load, it should be inside the library wherever needed.

peterszerzo:
This is addressed by #111.

PR #111, opened 2024-06-24
Support hosted triggers

peterszerzo:
@jakub-nlx also resolves your previous feedback about the `DOMContentLoaded` event being moved inside the SDK method, this will allow the setup script to be a bit simpler.

### DIFF
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -230,11 +230,15 @@ const resolveTriggers = async (
 
 // eslint-disable-next-line @typescript-eslint/promise-function-async
 const waitUntilDomContentLoaded = (): Promise<unknown> => {
-  return new Promise((resolve) => {
-    window.addEventListener("DOMContentLoaded", (event) => {
-      resolve(event);
+  if (document.readyState === "loading") {
+    return new Promise((resolve) => {
+      window.addEventListener("DOMContentLoaded", (event) => {
+        resolve();
+      });
     });
-  });
+  } else {
+    return Promise.resolve();
+  }
 };
 
 /**
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -187,9 +187,10 @@ export interface RunProps {
    */
   ui?: UiConfig;
   /**
-   * The triggers dictionary, downloaded from the Dialog Studio desktop app
+   * The triggers dictionary, downloaded from the Dialog Studio desktop app.
+   * If triggers are not provided, they will be fetched from the CDN.
    */
-  triggers: Triggers;
+  triggers?: Triggers;
   /**
    * Digression detection callback
    */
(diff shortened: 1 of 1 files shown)


packages/journey-manager/src/index.ts
```diff
@@ -187,9 +187,10 @@ export interface RunProps {
    */
   ui?: UiConfig;
   /**
-   * The triggers dictionary, downloaded from the Dialog Studio desktop app
+   * The triggers dictionary, downloaded from the Dialog Studio desktop app.
+   * If triggers are not provided, they will be fetched from the CDN.
    */
-  triggers: Triggers;
+  triggers?: Triggers;
```
jakub-nlx, line 193: Would it make more sense to pass in an explicit url rather than just null? That also gives them the opportunity to self-host if they so please?

packages/journey-manager/src/index.ts
```diff
@@ -209,14 +210,48 @@ function filterMap<X, Y>(
...
+  return triggersFromCdn;
+};
+
+// eslint-disable-next-line @typescript-eslint/promise-function-async
+const waitUntilDomContentLoaded = (): Promise<unknown> => {
+  return new Promise((resolve) => {
+    window.addEventListener("DOMContentLoaded", (event) => {
+      resolve(event);
+    });
+  });
```
jakub-nlx, line 237: See https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete.

```suggestion
  if (document.readyState === "loading") {
    return new Promise((resolve) => {
      window.addEventListener("DOMContentLoaded", (event) => {
        resolve();
      });
    });
  } else {
    return Promise.resolve();
  }
```

jakub-nlx (changes requested):
Closes https://github.com/nlxai/sdk/issues/93.

packages/journey-manager/src/index.ts
peterszerzo: I think we should check with the broader team whether the custom hosting functionality should be an option (has implications on whether we want to support users with that).

The hosting we will do is automatically configured under those specified URL's at all times, what is optional here is the triggers themselves (mostly for backwards-compatibility), which they can get in a two-liner if they want.

packages/journey-manager/src/index.ts
peterszerzo: Oh right on, thank you 🙏🙏🙏

jakub-nlx (approved):

packages/journey-manager/src/index.ts
```diff
@@ -209,14 +210,52 @@ function filterMap<X, Y>(
...
+  config: Config,
+  triggers?: Triggers,
+): Promise<Triggers> => {
+  if (triggers != null) {
+    return triggers;
+  }
+  const baseUrl =
+    config.dev ?? false
+      ? "https://triggers.dev.nlx.ai"
+      : "https://triggers.nlx.ai";
```
vlad31, line 223: It's `https://triggers.dev.mm.nlx.ai` and `https://triggers.mm.nlx.ai`

vlad31 (changes requested):

packages/journey-manager/src/index.ts
peterszerzo: @vlad31 oh good point, updated!

vlad31 (approved):

PR #115, opened 2024-07-02
Handle `appear` events

peterszerzo:
Also includes some refactoring that fixes a click but related to the `once` feature.

### DIFF
--- a/packages/journey-manager/src/index.ts
+++ b/packages/journey-manager/src/index.ts
@@ -70,7 +70,7 @@ interface LoadStep {
 }
 
 /**
- * Click step
+ * Step with additional query
  */
 export interface StepWithQuery {
   /**
@@ -90,6 +90,16 @@ export interface StepWithQuery {
    */
   urlCondition?: UrlCondition;
 }
+
+/**
+ * Step with query and found elements
+ */
+export type StepWithQueryAndElements = StepWithQuery & {
+  /**
+   * Elements found
+   */
+  elements?: HTMLElement[];
+};
 
 const debounce = (func: () => void, timeout = 300): (() => void) => {
   let timer: NodeJS.Timer | null = null;
@@ -124,6 +134,44 @@ const matchesUrlCondition = (urlCondition: UrlCondition): boolean => {
     return !url.includes(urlCondition.value);
   }
   return false;
+};
+
+const withElements = async (
+  steps: StepWithQuery[],
+): Promise<StepWithQueryAndElements[]> => {
+  const targets = await Promise.all(
+    steps
+      .filter(
+        ({ urlCondition }) =>
+          urlCondition == null || matchesUrlCondition(urlCondition),
+      )
+      .map(async (step) => {
+        try {
+          return {
+            ...step,
+            elements: await find(step.query),
+          };
+        } catch (e) {
+          return step;
+        }
+      }),
+  );
+  return targets;
+};
+
+const withElementsSync = (
+  steps: StepWithQuery[],
+): StepWithQueryAndElements[] => {
+  return filterMap(steps, (step) => {
+    if (step.urlCondition != null && !matchesUrlCondition(step.urlCondition)) {
+      return null;
+    }
+    const elements = getAll(step.query);
+    if (elements.length === 0) {
+      return null;
+    }
+    return { ...step, elements };
+  });
 };
 
 const localStorageKey = (conversationId: string): string =>
@@ -317,17 +365,18 @@ export const run = async (props: RunProps): Promise<RunOutput> => {
    * Handle load steps
    */
 
-  const loadSteps: LoadStep[] = Object.entries(triggers).reduce(
-    (prev: LoadStep[], [stepId, trigger]: [StepId, Trigger]) => {
+  const loadSteps: LoadStep[] = filterMap(
+    Object.entries(triggers),
+    ([stepId, trigger]: [StepId, Trigger]) => {
       if (trigger.event === "pageLoad") {
-        return [
-          ...prev,
-          { stepId, urlCondition: trigger.urlCondition, once: trigger.once },
-        ];
+        return {
+          stepId,
+          urlCondition: trigger.urlCondition,
+          once: trigger.once,
+        };
       }
-      return prev;
+      return null;
     },
-    [],
   );
 
   let previousUrl = window.location.toString();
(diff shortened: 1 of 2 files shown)


packages/journey-manager/src/index.ts
```diff
@@ -371,25 +404,24 @@ export const run = async (props: RunProps): Promise<RunOutput> => {
...
+          query: decode(trigger.query),
+          urlCondition: trigger.urlCondition,
+          once: trigger.once,
+        };
+        return [...prev, newEntry];
+      }
+      return prev;
+    },
+    [],
+  );
```
jakub-nlx, line 421: I'd suggest separating the generic filterMap out, it makes the logic easier to follow.

```suggestion
  function filterMap<A, B>(fn: A -> B | undefined | null, arr: A[]) : B[] {
     return arr.reduce((out, val) => {
        const res = fn(val);
        if (res != null) {
           out.push(res);
        }
        return res;
     }, []);
  }
  const appearSteps: StepWithQuery[] = filterMap(([stepId, trigger]: [StepId, Trigger]) => {
    if (trigger.event === "appear" && trigger.query != null) {
        const newEntry: StepWithQuery = {
          stepId,
          query: decode(trigger.query),
          urlCondition: trigger.urlCondition,
          once: trigger.once,
        }
    }, Object.entries(triggers));
```

packages/journey-manager/src/index.ts
```diff
@@ -126,6 +136,29 @@ const matchesUrlCondition = (urlCondition: UrlCondition): boolean => {
...
+    steps
+      .filter(
+        ({ urlCondition }) =>
+          urlCondition == null || matchesUrlCondition(urlCondition),
+      )
+      .map(async (step) => {
+        try {
+          return {
+            ...step,
+            elements: await find(step.query),
```
jakub-nlx, line 152: For appear on mutation observer, I'd imagine you want the sync version rather than waiting up to a second for the element to appear.

packages/journey-manager/src/index.ts
```diff
@@ -493,7 +525,27 @@ export const run = async (props: RunProps): Promise<RunOutput> => {
    * Change detection
    */
 
-  const documentObserver = new MutationObserver(() => {
+  const documentObserver = new MutationObserver((mutations) => {
+    // If any of the added nodes are inside matches on appear events, trigger those events
+    mutations.forEach(async (mutation) => {
+      const targets = await withElements(appearSteps);
```
jakub-nlx, line 531:
```suggestion
    const targets = await withElements(appearSteps);
    mutations.forEach(async (mutation) => {
```

Seems like a performance win, no?

packages/journey-manager/src/index.ts
```diff
@@ -493,7 +525,27 @@ export const run = async (props: RunProps): Promise<RunOutput> => {
...
+      targets.forEach(({ stepId, once, elements }) => {
+        if (
+          (elements ?? []).some((element) => {
+            if (
+              [...mutation.addedNodes].some((addedNode) => {
+                if (element.contains(addedNode)) {
+                  sendStep(stepId, once ?? false);
+                }
```
jakub-nlx, line 539: I'm not sure this is required at all. Don't we want basically the same logic like we do for URL events? If the DOM changed at all, and the query didn't succeed previously but now it did, then fire the appear event? Or perhaps more elaborately, we store the count of elements, then if the count is now larger than the old count we trigger the appear event?

packages/journey-manager/src/index.ts
peterszerzo: You already implemented one, I rewrote everything to use it.

packages/journey-manager/src/index.ts
peterszerzo: I don't see this the same way from a semantic perspective. If, by 'query succeeds', you mean both query and URL condition combined, then it can happen that the step fires even though no 'appear' took place, but purely from the technicality of the combined condition wasn't satisfied before but now is.

Practical example: we want to play something on voice when an error message first appears, as opposed to when the URL is finally compatible but the error had been around on the page for 2 minutes.

jakub-nlx: Well then pre-filter the array based on URL condition, then reset it when those change?

The above code as far as I read it just triggers when an element we are targeting has a  child inserted into the DOM. That could be for all sorts of reasons and doesn't necessarily mean it "appeared". For one, we are not checking its visibility, so it might not actually be visible to the user at all.

Also, why the `contains` check? That could trigger an element that simply has changeable children?

packages/journey-manager/src/index.ts
```diff
@@ -355,52 +404,43 @@ export const run = async (props: RunProps): Promise<RunOutput> => {
...
+        return {
           stepId,
           query: decode(trigger.query),
           urlCondition: trigger.urlCondition,
           once: trigger.once,
         };
-        return [...prev, newEntry];
       }
-      return prev;
+      return null;
```
jakub-nlx, line 418:
```suggestion
```

not necessary

packages/journey-manager/src/index.ts
peterszerzo: I had the condition backwards, just changed to `addedNode.contains(element)`. This way we make sure if the user selected an element, but that is inserted as a result of its parent being inserted, then the event still fires. Otherwise, the user has to target the exact element that has been added to the DOM.

Appear intentionally does not care about visibility. There is a separate 'enters viewport' that takes care of that, and they are intentionally separate because sometimes an error notification is not visible, but you want to react to it immediately anyway.

At this point I'm tempted to say if you have a different implementation in mind, feel free to suggest it.

packages/journey-manager/src/index.ts
peterszerzo: Explicit returns read clearer to me.

packages/journey-manager/src/index.ts
jakub-nlx: > I had the condition backwards, just changed to addedNode.contains(element).

OK, that I think alleviates most of my concern here.

> Appear intentionally does not care about visibility.

I believe that isn't strictly true, since the accessibility queries do care about elements being visible (at least in the they don't have `display: none` set sense).

jakub-nlx (approved):

packages/journey-manager/src/index.ts
peterszerzo: > I believe that isn't strictly true, since the accessibility queries do care about elements being visible (at least in the they don't have display: none set sense).

I'm ok with this edge case, but good to know for debugging.

PR #117, opened 2024-07-03
Journey manager intersection handling & bug fixes

peterszerzo:


packages/journey-manager/src/index.ts
```diff
@@ -458,6 +481,14 @@ export const run = async (props: RunProps): Promise<RunOutput> => {
         )
         .map((trigger) => ({ trigger, elements: getAll(trigger.query) }));
     }
+    if (eventType === "enterViewport") {
+      return enterViewportSteps
+        .filter(
+          ({ urlCondition }) =>
+            urlCondition == null || matchesUrlCondition(urlCondition),
+        )
+        .map((trigger) => ({ trigger, elements: getAll(trigger.query) }));
```
jakub-nlx, line 490: I'd extract the conditional and processing steps here. So something like:

```
const stepsForEvent = (
    eventType: ActiveTriggerEventType,
  ) : StepWithQuery[] => {
   switch(eventType) {
     case "click": return clickSteps;
     case "enterViewport": return enterViewportSteps;
  }
}

const findActiveTriggers = (
    eventType: ActiveTriggerEventType,
  ): ActiveTrigger[] => 
       stepsForEvent(eventType)
       .filter(
          ({ urlCondition }) =>
            urlCondition == null || matchesUrlCondition(urlCondition),
        )
        .map((trigger) => ({ trigger, elements: getAll(trigger.query) }));
```

packages/journey-manager/src/index.ts
```diff
@@ -549,14 +610,20 @@ export const run = async (props: RunProps): Promise<RunOutput> => {
...
-    const newUrl = window.location.toString();
-    if (newUrl !== previousUrl) {
-      handleLoadSteps();
+    if (props.ui?.highlights ?? false) {
+      setHighlights();
     }
-    previousUrl = newUrl;
+    observeIntersections();
+    // This timeout here seems to solve some timing issues on URL handling
+    setTimeout(() => {
```
jakub-nlx, line 618: Hm...

I'd like to understand a lot more about this. In my experience sprinkling `setTimeout` throughout your codebase is just a way to sweep bugs under the carpet.

What exact issues are you observing that this mitigates?

jakub-nlx (changes requested):

packages/journey-manager/src/index.ts
peterszerzo: If you take out the `setTimeout`, page load events for URL changes are completely messed up. For example: if I'm on `#tab1` and change the URL to `#tab2`, any page load event associated with `#tab2` will fire only when you leave `#tab2` to say `#tab3`. With the timeout, everything works reliably.

packages/journey-manager/src/index.ts
peterszerzo: Wound up with this solution:

```ts
  const findActiveTriggers = (
    eventType: ActiveTriggerEventType,
  ): ActiveTrigger[] => {
    const steps =
      eventType === "click"
        ? clickSteps
        : eventType === "enterViewport"
          ? enterViewportSteps
          : [];
    return steps
      .filter(
        ({ urlCondition }) =>
          urlCondition == null || matchesUrlCondition(urlCondition),
      )
      .map((trigger) => ({ trigger, elements: getAll(trigger.query) }));
  };

```

packages/journey-manager/src/index.ts
peterszerzo: I have an inkling it has to do with the timing/race condition of DOM mutations and URL changes, so with the timeout we are making sure that everything 'has settled' around the document change.

jakub-nlx (approved):

PR #118, opened 2024-07-03
allow nullable booleans

michaelglass:
follow up from https://github.com/nlxai/dialog-studio-frontend/pull/430#discussion_r1642663431

cc @jakub-nlx


jakub-nlx (approved):
Nice!

PR #121, opened 2024-07-03
test repo & ensure we build for every type

michaelglass:
addresses https://github.com/nlxai/sdk/issues/122

this is mergable. in a follow-up PR i'd like to add automated tests that load up a browser

addresses FRO-62 with some non-covering  integration tests

### DIFF
--- a/packages/chat-core/package.json
+++ b/packages/chat-core/package.json
@@ -2,6 +2,7 @@
   "name": "@nlxai/chat-core",
   "version": "0.1.36",
   "description": "NLX Chat SDK core",
+  "type": "module",
   "main": "lib/index.cjs",
   "module": "lib/index.esm.js",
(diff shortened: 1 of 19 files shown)


packages/chat-core/package.json
```diff
@@ -2,9 +2,9 @@
   "name": "@nlxai/chat-core",
   "version": "0.1.34",
   "description": "NLX Chat SDK core",
-  "main": "lib/index.esm.js",
-  "module": "lib/index.esm.js",
-  "commonjs": "lib/index.cjs.js",
+  "main": "lib/index.cjs",
```
michaelglass, line 5: I'm not sure how broad the compatibility is here, but if it's compatible, I'd suggest we roll this out to the rest of the packages. 

also note: main is cjs

jakub-nlx (approved):

PR #124, opened 2024-07-09
Expiration helpers

peterszerzo:
Expose helpers to figure out the current expiration timestamp of a conversation, allowing the user to handle it however they wish. Alternative to https://github.com/nlxai/sdk/pull/46.

```ts
const conversationHandler = nlxai.chatCore.create({ botUrl: "" });

conversationHandler.subscribe((responses) => {
  const timestamp = nlxai.chatCore.getCurrentExpirationTimestamp(responses);
  // Set timeouts, alerts etc.
});
```

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -945,6 +945,26 @@ export default function createConversation(
 }
 
 /**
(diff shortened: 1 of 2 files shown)


sam-trost (approved):
looks like the right stuff to me.

PR #119, opened 2024-07-03
unify rollup

michaelglass:
unifies rollup config across the application

PR #127, opened 2024-07-11
fix journey manager umd

michaelglass:
I haven't tested this extensively (across other packagers, for instance) but it appears to at least work for journey manager

### DIFF
--- a/packages/rollup-config-nlx/index.ts
+++ b/packages/rollup-config-nlx/index.ts
@@ -65,9 +65,7 @@ export default function buildconfig({
         commonjs(), // needed to bundle node modules that use cjs
         json(), // needed to fetch package version from package.json
         nodeResolve({ browser: true }), // bundles packages from node_modules
-        nodePolyfills({
-          include: "../../node_modules/**/*.js",
-        }),
+        nodePolyfills({ include: null }),
         terser(),
         replace({
           preventAssignment: true,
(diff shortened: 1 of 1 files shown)


packages/rollup-config-nlx/index.ts
```diff
@@ -65,9 +65,7 @@ export default function buildconfig({
         commonjs(), // needed to bundle node modules that use cjs
         json(), // needed to fetch package version from package.json
         nodeResolve({ browser: true }), // bundles packages from node_modules
-        nodePolyfills({
-          include: "../../node_modules/**/*.js",
-        }),
+        nodePolyfills({ include: null }),
```
michaelglass, line 68: another option here would be dropping the options altogether

packages/rollup-config-nlx/index.ts
```diff
@@ -65,9 +65,7 @@ export default function buildconfig({
         commonjs(), // needed to bundle node modules that use cjs
         json(), // needed to fetch package version from package.json
         nodeResolve({ browser: true }), // bundles packages from node_modules
-        nodePolyfills({
-          include: "../../node_modules/**/*.js",
```
michaelglass, line 69: before #119 this value was "../**/node_modules/**/*.js" (presumably taken from https://github.com/ionic-team/rollup-plugin-node-polyfills/issues/17#issuecomment-771803747).

I believe in 119 I changed it to `../..` to point to the proper `node_modules` path which may have caused this regression.

I still don't understand why `../**/node_modules` is relevant. I believe in the comment linked above, it's for a different monorepo setup where `node_modules` are symlinked into package directories.

In any case, the change to `../../` broke things. I thought I had confirmed that it worked by diffing umd output files but I certainly didn't test every package.

Do you understand what the previous `include` param was doing? If not I'd rather drop it altogether.

michaelglass:
used [this html](https://gist.github.com/michaelglass/a31226ab0cdbc81182959beb86845845) file to test the results of this change

In no way a comprehensive test!

peterszerzo (approved):

PR #133, opened 2024-07-22
Fix regex match

peterszerzo:


jakub-nlx (approved):
Oh yeah, what was I thinking!?

PR #134, opened 2024-07-23
Configuration for new features

peterszerzo:


packages/journey-manager/src/ui.tsx
```diff
@@ -76,17 +80,57 @@ export interface UiConfig {
...
    */
   onEscalation?: (config: { sendStep: Client["sendStep"] }) => void;
+  /**
+   * Escalation button label
+   */
+  escalationButtonLabel?: string;
+  /**
+   * Escalation confirmation
+   */
+  escalationConfirmation?: string;
```
jakub-nlx, line 90: I wonder if we should just make a button type and then use that for all of these?

```
interface Action<Payload> {
  buttonLabel?: string;
  confirmation?: string;
  icon?: string;
  handler?: (config: Payload) => void;
}

escalation?: Action<{sendStep: Client["sendStep"]}>;

end?: Action<{sendStep: Client["sendStep"]}>;

previousStep?: Action<{{
    sendStep: Client["sendStep"];
    triggeredSteps: Array<{ stepId: string; url: string }>;
  }>;

customButtons?: Array<Action<void>>
```

or some such?

peterszerzo: Yes I think we're converging towards what you're suggesting. This cannot quite happen yet because backwards-compatibility constraints but I will keep this in mind.

jakub-nlx: Well we could make the new api and the old and `console.warn` if you use the old, no?

peterszerzo: I even wonder if escalation and end should just fold into `customButtons` and those should just become `buttons`. There is nothing special about these types of functionality, if anything the SDK by design shouldn't care about end and escalation anyway, that is controlled inside Studio.

jakub-nlx: yeah if we don't have any particularly special logic than sure

peterszerzo: Yes there isn't. I'm going to try to spin it that way and if we get away with it then that will be the way forward.

michaelglass (approved):

PR #143, opened 2024-07-29
Adds a more robust box-shadow highlight strategy

jakub-nlx:
This now seems to work in all the scenarios I could think of... So in this branch this strategy is enabled by default. However, you can change this default by passing `highlightStrategy: 'overlay'` to the config.

### DIFF
--- a/packages/journey-manager/src/configuration.ts
+++ b/packages/journey-manager/src/configuration.ts
@@ -90,6 +90,11 @@ export interface UiConfig {
    * Render highlights
    */
   highlights?: boolean;
+
+  /**
+   * @hidden
+   */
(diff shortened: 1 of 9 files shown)


jakub-nlx:
Closes FRO-86.

jakub-nlx:
I should also mention that despite some claims I may have made to the contrary, I no longer believe that we could make this work very well for the dot ping look using box-shadow. But frankly I don't think the dot ping is possible to have in a way that won't break in all sorts of weird ways (and indeed that's why I implemented it in the desktop app using outlines rather than the pings originally despite what Kuba's designs were).

peterszerzo (approved):

PR #144, opened 2024-07-30
Journey Manager fixes

peterszerzo:
- fix control center confirmation alignment
- refactor and persist nudge state

### DIFF
--- a/packages/journey-manager/src/ui/components/index.tsx
+++ b/packages/journey-manager/src/ui/components/index.tsx
@@ -2,13 +2,7 @@
 /* eslint-disable jsdoc/require-jsdoc */
 import { type Client } from "@nlxai/multimodal";
 import { type FunctionComponent as FC, type ComponentChildren } from "preact";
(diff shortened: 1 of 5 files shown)


packages/journey-manager/src/ui/components/index.tsx
```diff
@@ -248,53 +248,102 @@ interface PinBubbleProps {
...
+    }
   }, [isOpen]);
 
   useEffect(() => {
-    if (hasBeenOpened || config.nudgeContent == null) {
-      setIsNudgeVisible(false);
+    saveNudgeState(conversationId, nudgeState);
+  }, [nudgeState, conversationId]);
+
+  useEffect(() => {
```
peterszerzo, line 319: @michaelglass reworked this logic a bit so that once dismissed, the nudge does not show up elsewhere. This also removes the 'hasBeenOpened' state.

jakub-nlx (approved):

PR #145, opened 2024-07-31
Fail-safe icon url

peterszerzo:


jakub-nlx (approved):
Another example of doing something complicated because our customers can't do something simple...

peterszerzo:
Yeah I hope we can strip it out sometime during a rewrite.

PR #150, opened 2024-08-07
update node

michaelglass:
updates node to newer minor version, pulls in newest nix release


peterszerzo (approved):

jakub-nlx:
The current LTS is 22... so we should probably try upgrading to that?

peterszerzo:
Fine by me :)

PR #152, opened 2024-08-15
Add confirmation to custom buttons

peterszerzo:
https://github.com/user-attachments/assets/1cd14973-9621-4215-88ea-691274bc9623


jakub-nlx (approved):
LGTM

PR #164, opened 2024-09-26
Export static icons

peterszerzo:
Export static icons so people can create custom buttons without having to host buttons for them.

Screenshot 2024-09-26 at 10 48 13 AM


jakub-nlx (dismissed):

michaelglass:
I don't understand how this fails and main doesn't. Diving in deeeper...

packages/website/src/components/Prototyping.tsx
```diff
@@ -1,5 +1,5 @@
 import { type FC, useRef, useEffect, useState } from "react";
-import { type Triggers, type Trigger } from "@nlxai/journey-manager";
```
michaelglass, line 2: previously, you were only importing types. When you started importing code here, server side rendering would get stuck on the web apis that were used in the journey-manager.

Another approach is exporting the icons in a separate file, but I didn't want to deal with creating another rollup config, etc.

michaelglass (approved):

vlad31 (approved):

PR #165, opened 2024-10-25
Handle uploads

peterszerzo:


jakub-nlx (dismissed):

packages/chat-widget/src/index.tsx
```diff
@@ -352,6 +357,75 @@ const isInputDisabled = (responses: Response[]): boolean => {
...
+      return;
+    }
+    setStatus("uploading");
+    try {
+      await fetch(props.upload.url, {
+        method: "PUT",
+        headers: {
+          "Content-Type": "image/jpeg",
+        },
+        body: new Blob([file], { type: file.type }),
```
jakub-nlx, line 406: I think this is unnecessary. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/File):

> A File object is a specific kind of [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob), and can be used in any context that a Blob can. In particular, the following APIs accept both Blobs and File objects:
> 
> ...
> - the [body](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#body) option to [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch)


So I think this can just be:

```suggestion
        body: file,
```

packages/chat-widget/src/index.tsx
```diff
@@ -352,6 +357,75 @@ const isInputDisabled = (responses: Response[]): boolean => {
   return new URLSearchParams(payload).get("nlx:input-disabled") === "true";
 };
 
+const findActiveUpload = (responses: Response[]): UploadUrl | null => {
+  const lastBotResponse = findLast(
```
jakub-nlx, line 361: Is there anyway there can be a partial upload and in the meantime another response comes in?

peterszerzo: Yes this is theoretically possible but I would say unlikely. Should probably change the state management to keep track of things a bit more granular, but I'm scared to do it now.

jakub-nlx: Maybe then at least add a TODO comment explaining this.

packages/chat-widget/src/index.tsx
```diff
@@ -255,18 +261,55 @@ const MessageGroups: FC<{
...
-              <C.MessageBody
-                dangerouslySetInnerHTML={{
-                  __html: marked(response.payload.text),
-                }}
-              />
-            </C.Message>
-          </C.MessageGroup>
-        );
+      if (response.type === "user") {
+        console.log(response);
```
jakub-nlx:
```suggestion
```

packages/chat-widget/src/index.tsx
```diff
@@ -255,18 +261,55 @@ const MessageGroups: FC<{
...
+                  />
+                ) : null}
+              </C.Message>
+              {uploadIds.map((uploadId) => {
+                const file = props.uploadedFiles[uploadId];
+                if (file != null) {
+                  return (
+                    <img
+                      key={uploadId}
+                      style={{ width: "100%" }}
```
jakub-nlx:
```suggestion
                      style={{ maxWidth: "100%" }}
```

in case its a tiny picture.

Also perhaps it should have rounded borders?

packages/chat-widget/src/index.tsx
```diff
@@ -255,18 +261,54 @@ const MessageGroups: FC<{
...
+                      src={URL.createObjectURL(file)}
```
jakub-nlx, line 302: This is much better than what I suggested 👍  I thought there was some better helper for this sort of thing...

packages/chat-widget/src/index.tsx
```diff
@@ -255,18 +261,54 @@ const MessageGroups: FC<{
...
+                      role="presentation"
```
jakub-nlx, line 303: Not sure this is exactly accurate since the picture is presumably a meaningful part of the conversation.

packages/chat-widget/src/index.tsx
```diff
@@ -255,18 +261,54 @@ const MessageGroups: FC<{
...
+                  />
+                ) : null}
+              </C.Message>
+              {uploadIds.map((uploadId) => {
+                const file = props.uploadedFiles[uploadId];
+                if (file != null) {
+                  return (
+                    <img
+                      key={uploadId}
+                      style={{ width: "100%", borderRadius: "10px" }}
```
jakub-nlx, line 301:
```suggestion
                      style={{ maxWidth: "100%", borderRadius: "10px" }}
```

in case the image is tiny

jakub-nlx (dismissed):

packages/chat-widget/src/index.tsx
peterszerzo: Added in 0f99847b43e8

packages/chat-widget/src/index.tsx
peterszerzo: Will add separately

packages/chat-widget/src/index.tsx
peterszerzo: I just don't know what alt attribute to add :/

jakub-nlx: Do you have access to the filename? I'd add that as the `alt`.

peterszerzo: Yes, added just that!

jakub-nlx (approved):

PR #171, opened 2025-01-04
Experimental touchpoint rendering

peterszerzo:
> Will most likely be abandoned in favor of #172 

This is a secret feature to allow the team to take touchpoint for a spin. When clicking 'open widget configurator' in Studio, they can add an additional `&touchpoint=true` query parameter, and get to use touchpoint instead of the old chat widget on the same page & the same bot.

Screenshot 2025-01-04 at 9 42 46 PM


peterszerzo:
Abandoned in favor of #172

PR #186, opened 2025-01-24
Fix responsiveness issues on touchpoint toggle

peterszerzo:


dtolb (approved):
Builds and runs on local. Approved ✅

PR #187, opened 2025-01-29
BE-1924: add stepTriggerDescription to /track endpoint

sam-trost:


ndrppnc (approved):

packages/voice-plus-core/src/index.ts
```diff
@@ -46,7 +46,7 @@ export const create = ({
     );
   }
   return {
-    sendStep: async (stepId: string, context?: Context) => {
+    sendStep: async (stepId: string, context?: Context, stepTriggerDescription?: string) => {
```
peterszerzo, line 49: @sam-trost what if we made the first argument take either `{ stepId: "", stepTriggerDescription: "" }` or just a string step ID, I'm a bit wary of users figuring out how to correctly write `sendStep("step_id", undefined, "step_description")`, and we're a bit better prepared if we need other fields in the future.

peterszerzo (changes requested):

packages/voice-plus-core/src/index.ts
jakub-nlx: Alternatively you could fidget with overloads to allow passing it as the second argument.

peterszerzo: @jakub-nlx thanks for the suggestion, I'm concerned people will mix them up, would rather just have named fields throughout.

sam-trost: I think we're fine to introduce a breaking change and increment major version on the sdk.

peterszerzo: It won't be breaking, you will be able to do `sendStep("my-step-id")` or `sendStep({ stepId: "my-step-id", stepTriggerDescription: "" })` and they will both work.

packages/voice-plus-core/src/index.ts
```diff
@@ -83,6 +89,13 @@ export const create = ({
   };
 };
 
+/**
+ * Step information, either a step ID as a single string or an object
+ */
+export type StepInfo =
+  | string
+  | { stepId: string; stepTriggerDescription?: string };
```
sam-trost, line 97: 👍

PR #189, opened 2025-02-03
Add icon support to launch button

peterszerzo:
Add launch icon and brand icon support to Touchpoint.

Fixes https://linear.app/nlxai/issue/FRO-614/add-touchpoint-logo-to-button-in-collapsed-mode.

https://github.com/user-attachments/assets/bbbe6c13-2427-46b3-a9fb-cf7d4b288831

### DIFF
--- a/packages/touchpoint-ui/src/App.tsx
+++ b/packages/touchpoint-ui/src/App.tsx
@@ -27,7 +27,6 @@ import ChatInput from "./components/ChatInput";
 import {
   type ColorMode,
   type WindowSize,
-  type LogoUrl,
   type ChoiceMessage,
   type Theme,
   type CustomModalityComponent,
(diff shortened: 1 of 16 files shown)


packages/touchpoint-ui/src/App.tsx
```diff
@@ -39,7 +39,19 @@ export interface Props {
   config: Config;
   windowSize?: WindowSize;
   colorMode?: ColorMode;
+  /**
+   * Logo URL
+   * @deprecated - use brandIcon and launchIcon intead
```
gampleman, line 44: We haven't released and already there's something deprecated? 🙄

packages/touchpoint-ui/src/App.tsx
```diff
@@ -39,7 +39,19 @@ export interface Props {
...
+   */
   logoUrl?: LogoUrl;
+  /**
+   * Icon used to display the brand in the chat header
```
gampleman, line 48:
```suggestion
   * URL to icon used to display the brand in the chat header
```

gampleman (approved):

packages/touchpoint-ui/src/App.tsx
peterszerzo: You're right, I'll take it out!

packages/touchpoint-ui/src/App.tsx
peterszerzo: Updated both places, thanks for the catch!

jakub-nlx (approved):

PR #195, opened 2025-02-17
Full-screen error page

peterszerzo:
Fixes https://linear.app/nlxai/issue/FRO-621/full-screen-error-page.


jakub-nlx (commented):
This is missing the

```css
@property --glow-opacity {
      syntax: "<percentage>";
      inherits: false;
      initial-value: 0%;
    }
```
declaration. 

This will cause the animation not to interpolate properly but to step between the different values.

PR #203, opened 2025-03-18
Voice in Touchpoint

peterszerzo:
Fixes https://linear.app/nlxai/issue/FRO-657/touchpoint-voice-frontend.

How to test / run locally:
- `npm install` top level
- `cd packages/touchpoint-ui`
- `npm run dev`

Edit the `packages/touchpoint-ui/index.html` file for application credentials and testing code.

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -1141,7 +1141,7 @@ export function createConversation(config: Config): ConversationHandler {
       return data;
     },
     terminateLiveKitCall: async () => {
-      const res = await fetch(`${applicationUrl}/terminateVoice`, {
+      const res = await fetch(`${fullApplicationHttpUrl()}/terminateVoice`, {
         method: "POST",
         headers: {
           ...(config.headers ?? {}),
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -501,6 +501,28 @@ export interface BotRequest {
 }
 
 /**
+ * Credentials to connect to a LiveKit channel
+ */
+export interface LiveKitCredentials {
+  /**
+   * LiveKit URL
+   */
+  url: string;
+  /**
+   * LiveKit room name
+   */
+  roomName: string;
+  /**
+   * LiveKit token
+   */
+  token: string;
+  /**
+   * LiveKit participant name
+   */
+  participantName: string;
+}
+
+/**
  * Helps link the choice to the specific message in the conversation.
  */
 export interface ChoiceRequestMetadata {
@@ -582,6 +604,19 @@ export interface ConversationHandler {
    * @param context - [Context](https://docs.studio.nlx.ai/workspacesettings/documentation-settings/settings-context-attributes) for usage later in the intent.
    */
   sendIntent: (intentId: string, context?: Context) => void;
+
+  /**
+   * Obtain LiveKit credentials to run the experience in voice.
+   * @internal
+   * @returns LiveKit credentials in promise form
+   */
+  getLiveKitCredentials: () => Promise<LiveKitCredentials>;
+
+  /**
+   * Terminate LiveKit call
+   * @internal
+   */
+  terminateLiveKitCall: () => Promise<unknown>;
 
   /**
    * Send a combination of choice, slots, and intent in one request.
@@ -711,6 +746,13 @@ export function createConversation(config: Config): ConversationHandler {
     conversationId: initialConversationId,
   };
 
+  const fullApplicationHttpUrl = (): string =>
+    `${applicationUrl}${
+      config.experimental?.completeBotUrl === true
+        ? ""
+        : `-${state.languageCode}`
+    }`;
+
   const setState = (
     change: Partial<InternalState>,
     // Optionally send the response that causes the current state change, to be sent to subscribers
@@ -822,23 +864,16 @@ export function createConversation(config: Config): ConversationHandler {
       }
     } else {
       try {
-        const res = await fetch(
-          `${applicationUrl}${
-            config.experimental?.completeBotUrl === true
-              ? ""
-              : `-${state.languageCode}`
-          }`,
-          {
-            method: "POST",
-            headers: {
-              ...(config.headers ?? {}),
-              Accept: "application/json",
-              "Content-Type": "application/json",
-              "nlx-sdk-version": packageJson.version,
-            },
-            body: JSON.stringify(bodyWithContext),
+        const res = await fetch(fullApplicationHttpUrl(), {
+          method: "POST",
+          headers: {
+            ...(config.headers ?? {}),
+            Accept: "application/json",
+            "Content-Type": "application/json",
+            "nlx-sdk-version": packageJson.version,
           },
-        );
+          body: JSON.stringify(bodyWithContext),
+        });
         if (res.status >= 400) {
           throw new Error(`Responded with ${res.status}`);
         }
@@ -1078,6 +1113,54 @@ export function createConversation(config: Config): ConversationHandler {
     },
     currentLanguageCode: () => {
       return state.languageCode;
+    },
+    getLiveKitCredentials: async () => {
+      const res = await fetch(`${fullApplicationHttpUrl()}/requestToken`, {
+        method: "POST",
+        headers: {
+          ...(config.headers ?? {}),
+          Accept: "application/json",
+          "Content-Type": "application/json",
+          "nlx-conversation-id": state.conversationId,
+          "nlx-sdk-version": packageJson.version,
+        },
+        body: JSON.stringify({
+          languageCode: state.languageCode,
+          conversationId: state.conversationId,
+          userId: state.userId,
+          requestToken: true,
+        }),
+      });
+      if (res.status >= 400) {
+        throw new Error(`Responded with ${res.status}`);
+      }
+      const data = await res.json();
+      if (data?.url == null) {
+        throw new Error("Invalid response");
+      }
+      return data;
+    },
+    terminateLiveKitCall: async () => {
+      const res = await fetch(`${applicationUrl}/terminateVoice`, {
+        method: "POST",
+        headers: {
+          ...(config.headers ?? {}),
+          Accept: "application/json",
+          "Content-Type": "application/json",
+          "nlx-conversation-id": state.conversationId,
+          "nlx-sdk-version": packageJson.version,
+        },
+        body: JSON.stringify({
+          languageCode: state.languageCode,
+          conversationId: state.conversationId,
+          userId: state.userId,
+        }),
+      });
+      if (res.status >= 400) {
+        throw new Error(`Responded with ${res.status}`);
+      }
+      // It has not been decided what will be sent here
+      return {};
     },
     subscribe,
     unsubscribe,
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -616,7 +616,7 @@ export interface ConversationHandler {
    * Terminate LiveKit call
    * @internal
    */
-  terminateLiveKitCall: () => Promise<unknown>;
+  terminateLiveKitCall: () => Promise<void>;
 
   /**
    * Send a combination of choice, slots, and intent in one request.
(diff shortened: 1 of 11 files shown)


packages/chat-core/src/index.ts
```diff
@@ -500,6 +500,28 @@ export interface BotRequest {
...
+  /**
+   * LiveKit token
+   */
+  token: string;
+  /**
+   * LiveKit participant name
+   */
+  participantName: string;
+}
+
```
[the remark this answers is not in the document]
peterszerzo: I'm a bit hesitant to set up a more abstract concept of voice credentials given that if we start supporting another provider, what you would do with the credentials would be substantially different. I'd say let's mark it as internal and see how things evolve.

packages/chat-core/src/index.ts
```diff
@@ -583,6 +605,19 @@ export interface ConversationHandler {
...
+   * @internal
+   * @returns LiveKit credentials in promise form
+   */
+  getLiveKitCredentials: () => Promise<LiveKitCredentials>;
+
+  /**
+   * Terminate LiveKit call
+   * @internal
+   */
+  terminateLiveKitCall: () => Promise<unknown>;
```
jakub-nlx, line 619:
```suggestion
  terminateLiveKitCall: () => Promise<void>;
```

it's a bit weird to have `unknown` in our own API.

packages/chat-core/src/index.ts
```diff
@@ -1079,6 +1114,54 @@ export function createConversation(config: Config): ConversationHandler {
...
+          languageCode: state.languageCode,
+          conversationId: state.conversationId,
+          userId: state.userId,
+          requestToken: true,
+        }),
+      });
+      if (res.status >= 400) {
+        throw new Error(`Responded with ${res.status}`);
+      }
+      const data = await res.json();
```
jakub-nlx, line 1137: I'd extract a fetch wrapper such as:

```ts
const post<T> = async (path : string, headers : Record<string, string>, body: any) : Promise<T> => {
   const res = await fetch(`${fullApplicationHttpUrl()}${path}`, {
        method: "POST",
        headers: {
          ...(config.headers ?? {}),
          Accept: "application/json",
          "Content-Type": "application/json",
          "nlx-sdk-version": packageJson.version,
          ...headers
        },
        body: JSON.stringify(body),
      });
      if (res.status >= 400) {
        throw new Error(`Responded with ${res.status}`);
      }
      return await res.json();
}
```

then use it everywhere.

packages/touchpoint-ui/src/types.ts
```diff
@@ -207,4 +207,8 @@ export interface TouchpointConfiguration {
    * @param handler - the conversation handler.
    */
   initializeConversation?: InitializeConversation;
+  /**
+   * Enable voice
+   */
+  voice?: boolean;
```
jakub-nlx, line 213: Is there no "voice only" touchpoint experience? If so, how do we foresee this API evolving?

packages/touchpoint-ui/src/voice.ts
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable jsdoc/require-jsdoc */
```
jakub-nlx, line 1: Can we somehow configure this rule only to apply to types that are actually visible to customers? 

...In fact, don't we already have tooling that does that? And if so, what's the value of even having this eslint rule in the first place?

packages/touchpoint-ui/src/voice.ts
```diff
@@ -0,0 +1,97 @@
...
+        const room = new Room();
+
+        setRoomState("pending");
+
+        const creds = await handler.getLiveKitCredentials();
+
+        const handleTrackSubscribed = (track: RemoteTrack): void => {
+          if (track.kind === Track.Kind.Audio) {
+            const element = track.attach();
+            document.body.appendChild(element);
```
jakub-nlx, line 60: We should attach this inside our shadow DOM, and not polute the customers DOM with this.

packages/touchpoint-ui/src/voice.ts
```diff
@@ -0,0 +1,97 @@
...
+          }
+        };
+
+        const handleIsSpeakingChanged = (speaking: boolean): void => {
+          setIsUserSpeaking(speaking);
+        };
+
+        await room.prepareConnection(creds.url, creds.token);
+        await room.localParticipant.setMicrophoneEnabled(true);
```
jakub-nlx, line 69: Does this ask for permission? I think it would make some sense to render something about that in the Touchpoint UI as (especially Chrome's) permission prompts can be quite subtle and the user could easily miss it (or instinctively dismiss and then have no idea next time what went wrong or how to fix it).

packages/touchpoint-ui/src/voice.ts
```diff
@@ -0,0 +1,97 @@
...
+        room.on(RoomEvent.TrackSubscribed, handleTrackSubscribed);
+        room.localParticipant.on(
+          ParticipantEvent.IsSpeakingChanged,
+          handleIsSpeakingChanged,
+        );
+        await room.connect(creds.url, creds.token, { autoSubscribe: true });
+        await room.startAudio();
+        setRoomState("active");
+      } catch (err) {
+        // TODO: error handling
```
jakub-nlx, line 79: Let's get at least some user feedback, even if it we just highlight the button in red or something.

packages/touchpoint-ui/src/voice.ts
```diff
@@ -0,0 +1,97 @@
...
+          console.warn(err);
+          // eslint-disable-next-line no-console
+          console.warn(err);
+        });
+      }
+    };
+    setup().catch((err) => {
+      // eslint-disable-next-line no-console
+      console.warn(err);
+    });
```
jakub-nlx, line 93: How can this `catch` ever be triggered if the whole body of `setup` is wrapped in a `try ... catch`?

jakub-nlx (changes requested):

packages/chat-core/src/index.ts
peterszerzo: In this case I find it to be the most honest way to type it, I'm trying to say 'something comes in the response, do not rely on it'.

packages/touchpoint-ui/src/types.ts
peterszerzo: I'm not sure actually, I found that enabling voice stuff has to rely on a user clicking a button, so showing a chat option before they consented to a voice interaction seems like a decent default pattern.

packages/touchpoint-ui/src/voice.ts
peterszerzo: I lost handle on how all of that works... this didn't use to be a problem but now with this vite-based project it somehow suddenly is.

packages/touchpoint-ui/src/voice.ts
peterszerzo: That's a question for Kuba, I will ask.

packages/touchpoint-ui/src/voice.ts
peterszerzo: It wouldn't be, this is just to make the linter happy because I suppose typescript can't possibly know statically that the promise will never reject.

packages/touchpoint-ui/src/voice.ts
peterszerzo: Yeah I actually have a few doubts on the permissions, like what if the user denies by accident and is not prompted for the permissions again...

packages/touchpoint-ui/src/voice.ts
peterszerzo: Actually I also wonder if it could just be played without attaching to the dom - thanks for bringing it up!

packages/chat-core/src/index.ts
peterszerzo: It's a good suggestion, I would just like to hold off on it until the dust settles on everything, right now it's full application URL one day, partial the next, don't think I have enough information to decide what the best abstraction is.

packages/touchpoint-ui/src/voice.ts
jakub-nlx: Usually in that case I think there is a little button in the nav bar where you can re-enable the permissions, but it's pretty subtle. If there's a hint in the UI to look for it, the percentage of users who will find it will be significantly higher than if they have no idea they are even supposed to be looking for it.

packages/chat-core/src/index.ts
jakub-nlx: yea

packages/chat-core/src/index.ts
```diff
@@ -1079,6 +1114,52 @@ export function createConversation(config: Config): ConversationHandler {
...
+      if (data?.url == null) {
+        throw new Error("Invalid response");
+      }
+      return data;
+    },
+    terminateLiveKitCall: async () => {
+      const res = await fetch(`${applicationUrl}/terminateVoice`, {
```
dtolb, line 1144: @peterszerzo  Should be `const res = await fetch(`${fullApplicationHttpUrl()}/terminateVoice`, {` instead (for now)

dtolb (changes requested):

packages/chat-core/src/index.ts
peterszerzo: Done!

packages/touchpoint-ui/src/voice.ts
```diff
@@ -82,11 +89,10 @@ export const useVoice = ({
...
           // eslint-disable-next-line no-console
           console.warn(err);
-          // eslint-disable-next-line no-console
-          console.warn(err);
         });
       }
     };
+    // This function will never throw, the catch is only there to make the linter happy
     setup().catch((err) => {
       // eslint-disable-next-line no-console
```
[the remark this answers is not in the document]
peterszerzo: Yesss great idea, done!

packages/touchpoint-ui/src/voice.ts
peterszerzo: Tracking in https://github.com/nlxai/sdk/issues/217, marking this chat as resolve.

packages/touchpoint-ui/src/types.ts
peterszerzo: @jakub-nlx is this any better / more future-proof? Let me know if you have alternatives.

Screenshot 2025-04-15 at 9 47 45 AM

jakub-nlx: I suppose the alternatives are:

1. What you suggest above. Is probably the most flexible, albeit a bit wonky in some ways.
2. `{ text? : boolean = true; voice?: boolean = false }` - has the downside that if the user passes `{ text: false, voice: false }` what do we do? 
3. Lean into modalities `{ modalities : { text: true, voice: false, customModality1: MyCustomModalityComponent } }`, as both text and voice are modalities, as is some future voice plus thingy.

Actually, speaking of Voice+, why is that limited to voice? Why can't you automate a website using a text conversation?

packages/chat-core/src/index.ts
jakub-nlx: Has this now settled, or is this still in the air?

packages/touchpoint-ui/src/components/Input.tsx
```diff
@@ -215,7 +233,31 @@ export const Input: FC<InputProps> = ({
...
+                  isUserSpeaking ? "shadow-[0_0_8px_var(--accent-20)]" : "",
+                )}
+                Icon={voiceActive ? MicOff : Mic}
+                label="Voice"
+                type={
+                  voiceActive
+                    ? roomState === "error"
+                      ? "error"
+                      : "activated"
+                    : "ghost"
```
jakub-nlx, line 250: _nit:_ I'd find it a bit easier to parse if this was extracted to a function and used proper conditionals. I struggle with the `?`/ `:` operator soup when it gets nested

packages/touchpoint-ui/src/voice.ts
jakub-nlx: Any more progress on this?

jakub-nlx (commented):
Another UX thing we should handle is to show some good feedback when the audio is playing, since the user may have their sound muted and could do with a reminder to unmute.

packages/touchpoint-ui/src/voice.ts
peterszerzo: Nooo it's outstanding...

packages/chat-core/src/index.ts
peterszerzo: If you ask me it's still too early

PR #213, opened 2025-04-08
Custom conversation initialization for touchpoint

peterszerzo:
Fixes https://linear.app/nlxai/issue/FRO-771/touchpoint-ui-set-context-on-open.

### DIFF
--- a/packages/touchpoint-ui/src/App.tsx
+++ b/packages/touchpoint-ui/src/App.tsx
@@ -109,15 +109,23 @@ const App = forwardRef<AppRef, Props>((props, ref) => {
     };
   }, [handler, setResponses]);
 
-  const initialWelcomeIntentSent = useRef<boolean>(false);
-
-  useEffect(() => {
-    if (!isExpanded || initialWelcomeIntentSent.current) {
+  const initializeConversation = useCallback(() => {
+    if (props.initializeConversation != null) {
+      props.initializeConversation(handler);
       return;
     }
-    initialWelcomeIntentSent.current = true;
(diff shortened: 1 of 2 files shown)


packages/touchpoint-ui/src/App.tsx
```diff
@@ -109,15 +109,23 @@ const App = forwardRef<AppRef, Props>((props, ref) => {
...
   }, [handler, setResponses]);
 
-  const initialWelcomeIntentSent = useRef<boolean>(false);
+  const initializeConversation = useCallback(() => {
+    if (props.initializeConversation != null) {
+      props.initializeConversation(handler);
+      return;
+    }
+    handler.sendWelcomeIntent();
+  }, [props.initializeConversation, handler]);
```
jakub-nlx, line 118: I'm thinking it would be somewhat cleaner if we handled default values for configuration a layer above the react application. That way we can probably avoid the `useCallback` and also just make this code a bit simpler.

jakub-nlx (changes requested):

packages/touchpoint-ui/src/App.tsx
peterszerzo: That would come at the expense of the application no longer working standalone as a React component, I realize the advantage is a bit speculative but I'm still wondering if we might hang on to that given that this isn't quite so complicated yet. What do you think?

peterszerzo: That's fair, let's do it that way!

PR #215, opened 2025-04-08
fix markdown anchor links

michaelglass:
I _guess_ this was a regression introduced in https://github.com/nlxai/sdk/commit/7be6c3d5c198069cb3b9d2a4553733b739e6bdb8 but I have no confirmed that.

The markdown code had named anchor tags that were stripped out.

After this commit, I believe markdown anchors work again.

### DIFF
--- a/packages/website/src/components/PageContent.tsx
+++ b/packages/website/src/components/PageContent.tsx
@@ -71,7 +71,6 @@ export const PageContent: FC<{ md: string; className?: string }> = ({
             const name: string | undefined = (props as unknown as any).name;
             // eslint-disable-next-line react/prop-types
             if (props.href == null && name != null) {
-              // eslint-disable-next-line react/prop-types
               return <a id={name}>{props.children}</a>;
             }
             // TODO: when using a react-router redirect, scroll to heading ID if available
--- a/packages/website/src/components/PageContent.tsx
+++ b/packages/website/src/components/PageContent.tsx
@@ -66,11 +66,12 @@ export const PageContent: FC<{ md: string; className?: string }> = ({
         rehypePlugins={[rehypeRaw, rehypeSlug]}
         components={{
           a(props) {
-            // JSDOC output contains markup such as <a name="interfacespropsmd"></a>, so the `name` attribute shows up even if the
+            // `concat-md` output contains markup such as <a name="interfacespropsmd"></a> as anchor targets. However, react types don't expect the (basically deprecated) `name` attribute, hence the `as unknown as any` hack.
             // markdown package typings expect it not to.
             const name: string | undefined = (props as unknown as any).name;
             // eslint-disable-next-line react/prop-types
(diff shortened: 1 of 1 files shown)


packages/website/src/components/PageContent.tsx
```diff
@@ -66,9 +66,9 @@ export const PageContent: FC<{ md: string; className?: string }> = ({
       components={{
         a(props) {
           // eslint-disable-next-line react/prop-types
-          if (props.href == null) {
+          if (props.href == null && props.name != null) {
             // eslint-disable-next-line react/prop-types
-            return <a href={props.href}>{props.children}</a>;
+            return <a id={props.name}>{props.children}</a>;
```
michaelglass, line 71: @dtolb I wasn't sure why you were passing a null href through. Is there something here I was missing?

🙏

jakub-nlx (approved):

packages/website/src/components/PageContent.tsx
peterszerzo: @michaelglass how are regular links supposed to work here? Or any link for that matter, without an href attribute. You're probably right about the regression, I'm afraid we only tested them for regular markdown links from elsewhere on the page (not the generated docs).

peterszerzo: Oh never mind I understand now...

peterszerzo (commented):
@michaelglass @jakub-nlx added some comments etc. to explain what needed to be done to keep typescript happy.

peterszerzo (approved):

packages/website/src/components/PageContent.tsx
```diff
@@ -66,14 +66,16 @@ export const PageContent: FC<{ md: string; className?: string }> = ({
         rehypePlugins={[rehypeRaw, rehypeSlug]}
         components={{
           a(props) {
+            // JSDOC output contains markup such as <a name="interfacespropsmd"></a>, so the `name` attribute shows up even if the
+            // markdown package typings expect it not to.
+            const name: string | undefined = (props as unknown as any).name;
```
michaelglass, line 71: 🙏

packages/website/src/components/PageContent.tsx
```diff
@@ -66,14 +66,16 @@ export const PageContent: FC<{ md: string; className?: string }> = ({
...
             // eslint-disable-next-line react/prop-types
-            if (props.href == null) {
-              // eslint-disable-next-line react/prop-types
-              return <a href={props.href}>{props.children}</a>;
+            if (props.href == null && name != null) {
+              return <a id={name}>{props.children}</a>;
```
michaelglass, line 74:
```suggestion
              // redefine `name` as `id` as suggested in the MDN for anchor targets.
              return <a id={name}>{props.children}</a>;
```

michaelglass:
@peterszerzo  / @jakub-nlx regression test?

PR #222, opened 2025-04-22
Fix touchpoint paragraph spacing

peterszerzo:
Make sure there's spacing between multiple paragraphs:
 
Screenshot 2025-04-21 at 22 32 26

### DIFF
--- a/packages/touchpoint-ui/src/components/Messages.tsx
+++ b/packages/touchpoint-ui/src/components/Messages.tsx
@@ -187,7 +187,7 @@ export const Messages: FC<MessagesProps> = ({
                   return (
                     <div key={messageIndex} className="text-base">
                       <div
-                        className="pr-10"
+                        className="pr-10 space-y-6"
                         dangerouslySetInnerHTML={{
                           __html: marked(message.text),
                         }}


packages/touchpoint-ui/src/components/Messages.tsx
```diff
@@ -187,7 +187,7 @@ export const Messages: FC<MessagesProps> = ({
                   return (
                     <div key={messageIndex} className="text-base">
                       <div
-                        className="pr-10"
+                        className="pr-10 space-y-6"
```
jakub-nlx, line 190: We should figure out a way to use the [fancy new `lh` units to do this](https://www.trevorlasn.com/blog/css-lh-and-rlh-units), as it seems to lead to particularly nice typography with great flexibility (for instance if we let the users change font-size and/or line-height, they'd still get nice paragraph spacing that way).

jakub-nlx (approved):

packages/touchpoint-ui/src/components/Messages.tsx
peterszerzo: Oh cool 😍

PR #226, opened 2025-05-05
Bidirectional configuration

peterszerzo:


packages/chat-core/src/index.ts
```diff
@@ -1184,6 +1278,17 @@ export function createConversation(config: Config): ConversationHandler {
...
       botRequestOverride = val;
     },
+    addEventListener: (event, listener) => {
+      eventListeners[event] = [...eventListeners[event], listener];
+    },
+    removeEventListener: (event, listener) => {
+      eventListeners[event] = eventListeners[event].filter(
+        (l) => l !== listener,
+      );
+    },
```
jakub-nlx, line 1288: Perhaps we should inherit from `EventTarget` instead?

packages/touchpoint-ui/src/types.ts
```diff
@@ -218,5 +218,5 @@ export interface TouchpointConfiguration {
   /**
    * Controls the ways in which the user can communicate with the application. Defaults to `"text"`
    */
-  input?: "text" | "voice" | "textAndVoice";
+  input?: "text" | "voice" | "textAndVoice" | "experimental";
```
jakub-nlx, line 221: Is `experimental` the best we can come up with? Could we not do something like `experimentalBidirectional`?

packages/voice-plus-web/src/context.ts
```diff
@@ -0,0 +1,90 @@
...
+      placeholder: element.placeholder,
+      value: element.value,
+    };
+  } else if (element instanceof HTMLSelectElement) {
+    const accessibleName = computeAccessibleName(element) ?? "";
+    return {
+      name: accessibleName === "" ? element.name : accessibleName,
+      description: computeAccessibleDescription(element),
+      type: element.type,
+      value: element.value,
```
jakub-nlx, line 56: Shouldn't the context contain what all the options are?

packages/voice-plus-web/src/context.ts
```diff
@@ -0,0 +1,90 @@
...
+  throw new TypeError("Unsupported element type");
+};
+
+/**
+ * Analyze page forms
+ * @returns pageForms
+ */
+export const analyzePageForms = (): PageForms => {
+  const interactiveNodes = Array.from(
+    document.querySelectorAll("form input, form textarea, form select"),
```
jakub-nlx, line 68: Why do these need to be in `<form>` elements?

packages/voice-plus-web/src/context.ts
```diff
@@ -0,0 +1,90 @@
...
+    index,
+    id: `${element.tagName.toLowerCase()}-${index}`,
+    element,
+  }));
+
+  const formElements: Record<string, Element> = {};
+
+  elementInfo.forEach(({ id, element }) => {
+    formElements[id] = element;
+  });
```
jakub-nlx, line 81: If we're not serializing these, a `Map` would probably be better.

jakub-nlx (changes requested):

packages/touchpoint-ui/src/types.ts
peterszerzo: I'm happy to call it `"experimentalVoicePlus"`, bidirectional shouldn't be mentioned in public API.

packages/voice-plus-web/src/context.ts
peterszerzo: I don't know, this is copied over from the other PR, pretty sure we just blindly took over Andrei's work and added the accessibility API to it. Happy to remove.

packages/chat-core/src/index.ts
peterszerzo: I'm not against it, I just wonder if we can type it nicely in that case, or throw accurate types on top of it even if the implementation relies on `EventTarget`.

jakub-nlx: I think the latter should work.

peterszerzo: Nice, I'll look into it - it's worth it because then we can set up an alternative to `subscribe` and `unsubscribe`  that use events we can name nicely.

peterszerzo: Opened https://github.com/nlxai/sdk/issues/235, will handle in a different PR.

packages/voice-plus-web/src/context.ts
peterszerzo: Yes probably, I would say this should be in scope of refinements later on.

packages/voice-plus-web/src/context.ts
peterszerzo: Removed!

packages/voice-plus-web/src/context.ts
peterszerzo: Added a comment about it, I'm not super sure this code will stick around.

jakub-nlx (approved):

PR #234, opened 2025-05-13
Voice improvements

peterszerzo:
- set up prepare experience.
- add proper reset functionality.
- mic and speakers mute/unmute.


packages/touchpoint-ui/src/components/FullscreenVoice.tsx
```diff
@@ -1,52 +1,179 @@
...
+              )}
+            >
+              {soundCheck.micAllowed ? <Mic /> : <MicOff />}
+            </span>
+            <input
+              className="p-2 rounded-inner bg-primary-5 w-full flex-grow text-primary-80"
+              value={soundCheck.micNames[0]}
+              placeholder="Mic not found"
+              disabled
+            />
```
jakub-nlx, line 61: Would this not make more sense as a select? Or is the just placeholder UI?

packages/touchpoint-ui/src/components/VoiceMini.tsx
```diff
@@ -0,0 +1,104 @@
...
+  if (roomState === "pending") {
+    return (
+      <Container>
+        <div className="flex items-center p-8">
+          <span className="w-6 h-6 block text-accent">
+            <LoaderAnimation />
+          </span>
+        </div>
+      </Container>
+    );
```
jakub-nlx, line 75: Wouldn't loading make sense in the compact container as well?

packages/touchpoint-ui/src/voice.ts
```diff
@@ -56,79 +124,80 @@ export const useVoice = ({
...
+        setIsUserSpeaking(hasLocal);
+      };
+
+      const handleTrackSubscribed = (track: RemoteTrack): void => {
+        if (track.kind === Track.Kind.Audio) {
+          const element = track.attach();
+          element.play().catch((err) => {
+            // eslint-disable-next-line no-console
+            console.warn(err);
+          });
```
jakub-nlx, line 153:
```suggestion
          void element.play();
```

should work too, no?

packages/touchpoint-ui/src/voice.ts
```diff
@@ -56,79 +124,80 @@ export const useVoice = ({
...
+      const room = new Room();
+      roomRef.current = room;
+
+      await room.prepareConnection(creds.url, creds.token);
+      room.on(RoomEvent.TrackSubscribed, handleTrackSubscribed);
+      room.on(RoomEvent.ActiveSpeakersChanged, handleActiveSpeakersChanged);
+      room.localParticipant.on(
+        ParticipantEvent.IsSpeakingChanged,
+        handleIsSpeakingChanged,
+      );
```
jakub-nlx, line 171: I think I'd find this a bit easier to read if the functions here were inlined.

packages/touchpoint-ui/src/components/FullscreenVoice.tsx
peterszerzo: Yeah it doesn't work, it's always disabled and either shows the likely audio output or the placeholder.

packages/chat-core/src/index.ts
```diff
@@ -564,6 +564,36 @@ export type BotRequestOverride = (
...
+export type ConversationHandlerEvent = "voicePlusCommand";
+
+/**
+ * Dictionary of handler methods per event
+ */
+export interface EventHandlers {
+  /**
+   * Voice+ command event handler
+   */
+  voicePlusCommand: (payload: any) => void;
```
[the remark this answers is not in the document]
peterszerzo: I would like to see first what the NLU comes up with as we prototype features.

packages/touchpoint-ui/src/components/VoiceMini.tsx
peterszerzo: Do you mean adding the word `Loading...`

packages/chat-core/src/index.ts
```diff
@@ -564,6 +564,36 @@ export type BotRequestOverride = (
   appendBotResponse: (res: BotResponsePayload) => void,
 ) => void;
 
+/**
+ * Voice+ context, type to be defined
+ */
+export type VoicePlusContext = any;
+
```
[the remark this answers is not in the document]
peterszerzo: It's a bit awkward because this would introduce a `voice-plus-web` dependency unless we were willing to duplicate. But it would also change a lot the next few weeks.

packages/touchpoint-ui/src/voice.ts
peterszerzo: I hear you, would just want to wait out how this will change to make sure it works reliably.

packages/touchpoint-ui/src/components/VoiceMini.tsx
jakub-nlx: No as in using the CompactContainer component

peterszerzo: Oh sorry I understand what you mean now - I don't think so, the width shouldn't change quite yet given that it could error out, in which case it went from wide to narrow to wide again.

jakub-nlx (approved):

PR #237, opened 2025-05-15
Adds touchpoint to touchpoint

dtolb:
Updates the website package to install and use touchpoint itself. 

Key items

* Launching touchpoint HTML snippet in markdown
* Driving webpage actions / navigation via touchpoint modalities
* Touchpoint context for website

Known issue

* Launching "try it now" touchpoint to smartly replace primary touchpoint instance
* Voice toggle


packages/website/src/components/Touchpoint.tsx
```diff
@@ -0,0 +1,134 @@
...
+        );
+        navigate(redirectPayload.targetUrl);
+      } else {
+        console.warn(
+          "RedirectPage modality received but targetUrl is missing or not a string:",
+          redirectPayload,
+        );
+      }
+    }
+  };
```
dtolb, line 51: There's a better way to do this. Works for now though

peterszerzo (changes requested):
- should use tailwind classes as oppose to style attributes.
- directly importing `touchpoint-ui` will not work with server rendering, this is why we currently have to [import dynamically](https://github.com/nlxai/sdk/blob/5228269b7d6c9c9e69ccd8a5fc838c6840be985f/packages/website/src/content/02-01-touchpoint-ui-try-live.tsx#L139). While this is solvable, keep in mind it will only work if `npm run render-static` from `packages/website` runs without errors.
- (sorry to be blunt and vague, I'm short on time) I would like to look into a generally simpler approach here.

packages/website/src/content/04-03-touchpoint-Buttons.md
```diff
@@ -11,7 +11,11 @@
 
 ### Text Button Example
 
-<img src="/images/Touchpoint-Text-Button.png" alt="Text Button with Left Arrow Icon" style="max-width: 40%;">
+<div class="launch-touchpoint-button"
```
peterszerzo, line 14: We should make this into a React component-looking thing `<TouchpointTrigger title="" label="" description="" context={{}} />`, this would make rendering logic simpler.

peterszerzo (changes requested):

PR #241, opened 2025-05-16
Add mermaid chart support

peterszerzo:
It's frivolous but it's fun...

CleanShot 2025-05-16 at 15 25 52@2x


jakub-nlx (changes requested):
Do we have a way to ensure that the mermaid charts fit our overall design?

dtolb:
> Do we have a way to ensure that the mermaid charts fit our overall design?

@jakub-nlx , mermaid supports [theming](https://mermaid.js.org/config/theming.html) that can be added site-wide

PR #244, opened 2025-05-19
generate userId

dtolb:


packages/website/src/content/02-01-touchpoint-ui-try-live.tsx
```diff
@@ -138,8 +138,15 @@ export const Content: FC<unknown> = () => {
     // Import has to happen dynamically after mount because the bundle has an issue with server rendering at the moment
     import("@nlxai/touchpoint-ui/lib/index.js")
       .then(async ({ create }) => {
+        const touchpointConfig = {
+          ...config,
+        };
+        if (config.userId === undefined || config.userId === "") {
+          touchpointConfig.userId = "TemporaryUserId";
```
ndrppnc, line 145: Hardcoded ideas are a bad idea. Should be a UUID

ndrppnc (changes requested):

packages/website/src/content/02-01-touchpoint-ui-try-live.tsx
dtolb: Extracted to a new function and add a UUID with `TemporaryUserId` prefix. 

The `userId` in this function is generated by the dev docs and added to the config at launch. I think a prefix `TemporaryUserId` (or similar) is nice to have as then analytics and debug-events will show the user where it's temporary and not something they added. 

Easy to remove / change as well.

jakub-nlx (approved):

PR #245, opened 2025-05-20
Adds context to voice in touchpoint

jakub-nlx:
You'd add an `initialContext` parameter to the touchpoint config.

### DIFF
--- a/packages/chat-core/src/index.ts
+++ b/packages/chat-core/src/index.ts
@@ -640,7 +640,7 @@ export interface ConversationHandler {
    * @internal
    * @returns LiveKit credentials in promise form
    */
-  getLiveKitCredentials: () => Promise<LiveKitCredentials>;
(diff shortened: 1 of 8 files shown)


jakub-nlx:
@dtolb Yes! In fact that initialize function will receive the context you pass there as a second argument (just to make life easier for structuring code).

dtolb:
@jakub-nlx , can you give an example where I'd use both context and init function ?

jakub-nlx:
I think the only reason you'd want to use both is if things happen to be structured that way in your codebase.

I prefer this duplication over having a context field that only works for voice. The init function is more powerful in that it allows you to do more than just send the welcome intent, but the context I suppose is simpler.

jakub-nlx:
We could go the route of typing the config object better making sure you can't pass options that aren't relevant with the other options (such as an init function in voice mode), but I'm not convinced the extra complexity is worth it. But it might be worth mentioning for the init function docs that it only works in text mode...

choso (approved):

PR #248, opened 2025-05-27
updates the try-live to pull input from query parameters

dtolb:
updates the try-live to pull input from query parameters.. 

Allows for users to share configurations with other users and pull input = voice, voiceMini, text through to url query parameters


dtolb:
> This already happens - that's what `useUrlState` does.

Then i think there's a bug :( 

https://developers.nlx.ai/touchpoint-ui-try-live?input=voice

PR #249, opened 2025-05-29
Re-enable mute-unmute

peterszerzo:
I had taken these out because I wasn't sure they were causing voice issues, but turns out those issues were on the backend.


jakub-nlx (approved):

jakub-nlx:
Fixes FRO-980.

PR #256, opened 2025-06-03
Custom language for touchpoint-ui code

peterszerzo:
When using markdown code snippets tagged `touchpointui`, these will now render and environment toggle.


packages/website/src/components/PageContent.tsx
```diff
@@ -55,6 +57,80 @@ export const Prose: FC<{ children: ReactNode; className?: string }> = ({
...
 );
 
+type Env = "html" | "js";
+
+const touchpointUiImports = ["create", "React", "html"];
+
+/**
+ * Update this manually according to which version is considered ready for outside use
+ */
+const touchpointUiVersion = "1.0.5-alpha.10";
```
jakub-nlx, line 67: Should we just read this from package.json?

packages/website/src/components/PageContent.tsx
```diff
@@ -55,6 +57,80 @@ export const Prose: FC<{ children: ReactNode; className?: string }> = ({
...
+`;
+};
+
+const Code: FC<{
+  children: ReactNode;
+  className?: string;
+}> = ({ children, className }) => {
+  const language = /language-(\w+)/.exec(className ?? "")?.[1];
+  const lines = String(children).replace(/\n$/, "");
+  const [env, setEnv] = useState<Env>("html");
```
jakub-nlx, line 91: It would make sense if this toggle was global rather than controlled by any individual component.

peterszerzo: That would save clicks but I'm not sure if it would be clearer, usually with any tabs you would expect to influence the widget you're currently interacting with.

packages/website/src/components/PageContent.tsx
peterszerzo: I don't think it's easy, it would have to be the package.json of a different package, unless we made that package re-export its version.

jakub-nlx: LGTM.

jakub-nlx: Although, isn't this running in node? Can we not just `File.read` the relevant package.json?

PR #257, opened 2025-06-03
Add dom content loaded check into package

peterszerzo:


jakub-nlx:
I understood that. But why do we need the DOM to be ready in order to instantiate touchpoint?

jakub-nlx:
Also I think we were doing some of this stuff for Voice+ where we were interested in the user's DOM, so having it be complete seems important. But for touchpoint I'm not sure it will make any difference.

PR #263, opened 2025-06-06
Moves analyzePageForms to touchpoint for now

jakub-nlx:
Makes it easy to import and BiDI V+ at the moment only really works via touchpoint anyway.


dtolb (approved):
Very handy - thank you!

PR #266, opened 2025-06-08
feat: support passing room data to frontend

sam-trost:


packages/touchpoint-ui/src/voice.ts
```diff
@@ -191,6 +194,19 @@ export const useVoice = ({
         disconnect();
       });
 
+      // Handle incoming data from the room/agent
+      room.on(RoomEvent.DataReceived, (payload, participant) => {
+        try {
+          const data = JSON.parse(new TextDecoder().decode(payload));
+          console.log("Received data from LiveKit room:", data, "from participant:", participant?.identity);
```
sam-trost, line 201: feel free to remove these console logs...

packages/touchpoint-ui/src/voice.ts
```diff
@@ -29,6 +29,7 @@ interface VoiceHookReturn {
   isUserSpeaking: boolean;
   isApplicationSpeaking: boolean;
   soundCheck: null | SoundCheck;
+  roomData: any;
```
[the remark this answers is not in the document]
peterszerzo: @jakub-nlx I just found out that this is for voice modalities, which we have to add generics for anyway so they can propagate through a custom touchpoint implementation.

packages/touchpoint-ui/src/voice.ts
```diff
@@ -134,6 +134,7 @@ export const useVoice = ({
       audioElementRef.current.pause();
       audioElementRef.current = null;
     }
+    setRoomData(null);
   }, []);
```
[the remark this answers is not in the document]
peterszerzo: I added this

packages/chat-core/src/index.ts
```diff
@@ -104,13 +104,19 @@ export interface BotResponsePayload {
...
    * If the node is set to send context, the whole context associated with the conversation.
    */
   context?: Context;
 }
 
+/**
+ * Modalities data
+ * TODO: make it generic
+ */
+export type Modalities = Record<string, any>;
```
jakub-nlx, line 118: We should rename this to something like `ModalitiesPayload` or some such, I thought this was the custom modalities components.

packages/touchpoint-ui/index.html
```diff
@@ -89,20 +89,96 @@ <h1>Loreum ipsum dolor sit amet</h1>
...
-          userId: "1234-5678",
-          languageCode: "en-US",
+      import { create, html, Icons } from "/src/index.tsx";
+
+      // Example custom modality component for voice
+      const ExampleCard = ({ data, conversationHandler }) => {
+        return html`
+          <div className="bg-accent rounded-inner p-4 text-white shadow-lg">
+            <div className="flex items-center gap-2 mb-2">
+              <${Icons.Touchpoint} className="w-5 h-5" />
```
jakub-nlx, line 99: I thought the `html` helper included the components by default? Can't we just do:

```suggestion
              <Icons.Touchpoint className="w-5 h-5" />
```

jakub-nlx (commented):
What about voice mini? Should that have some pop-up version of these?

packages/chat-core/src/index.ts
peterszerzo: @jakub-nlx could I run two more names by you and you pick one (including your suggestion):
- `ModalitiesPayload`
- `ModalityPayloads`
- `PayloadsByModality`

packages/touchpoint-ui/index.html
peterszerzo: Actually per definition it would just be `<Touchpoint`, but maybe I made a mistake there? https://github.com/nlxai/sdk/blob/08064a60b11f139c5ef74152140deda1e5b8ce00/packages/touchpoint-ui/src/index.tsx#L58

packages/chat-core/src/index.ts
jakub-nlx: Any of those would do

jakub-nlx: `ModalityPayloads` is probably my favorite if its all the same to you

packages/touchpoint-ui/index.html
peterszerzo: I'm going to leave it like this for now so at least if we change it typescript catches it.

peterszerzo (approved):

peterszerzo:
> What about voice mini? Should that have some pop-up version of these?

@jakub-nlx I opened https://github.com/nlxai/sdk/issues/274 to cover it.

jakub-nlx (dismissed):

packages/chat-core/src/index.ts
```diff
@@ -104,13 +104,19 @@ export interface BotResponsePayload {
...
+  modalities?: ModalityPayloads;
   /**
    * If the node is set to send context, the whole context associated with the conversation.
    */
   context?: Context;
 }
 
+/**
+ * Modalities data
+ * TODO: make it generic
```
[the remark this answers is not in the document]
peterszerzo: I took it out, good point, we know we can improve types!

PR #275, opened 2025-06-12
Improve modality type

jakub-nlx:
The type variable is useless to us, but will allow users to strictly type their own components. So as a user I can now do:

```typescript 

import { type CustomModalityComponent } from "@nlxai/touchpoint-ui";

export type MyModality = {
  foo: string;
  bar: number[];
};

export const MyModalityHandler : CustomModalityComponent<MyModality> = ({data, conversationHandler}) => {
   return <div>{data.fo // <-- foo will be autocompleted here
}
```

But the new type will also support class based components (because why not? we're selling to enterprise after all).

Also before we didn't show the `conversationHandler` in the type.

### DIFF
--- a/packages/chat-widget/src/index.tsx
+++ b/packages/chat-widget/src/index.tsx
@@ -245,7 +245,13 @@ const MessageGroups: FC<{
                   );
                   return null;
                 }
-                return <Component key={key} data={value} />;
+                return (
+                  <Component
+                    key={key}
+                    data={value}
+                    conversationHandler={props.chat.conversationHandler}
+                  />
+                );
               },
             )}
           </C.MessageGroup>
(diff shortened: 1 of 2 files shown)


peterszerzo:
> The type variable is useless to us, but will allow users to strictly type their own components. So as a user I can now do:
> 
> ```ts
> import { type CustomModalityComponent } from "@nlxai/touchpoint-ui";
> 
> export type MyModality = {
>   foo: string;
>   bar: number[];
> };
> 
> export const MyModalityHandler : CustomModalityComponent<MyModality> = ({data, conversationHandler}) => {
>    return <div>{data.fo // <-- foo will be autocompleted here
> }
> ```
> 
> But the new type will also support class based components (because why not? we're selling to enterprise after all).
> 
> Also before we didn't show the `conversationHandler` in the type.

peterszerzo (changes requested):
I love it, let's add it to touchpoint-ui as well (chat-widget is deprecated)

jakub-nlx:
Oh we have 2 copies of this component... sorry I got confused.

peterszerzo (approved):

PR #279, opened 2025-06-12
Closable touchpoint mini

peterszerzo:
- fixes https://github.com/nlxai/sdk/issues/274.
- fixes https://linear.app/nlxai/issue/PRO-3736/cant-close-touchpoint-mini.

https://github.com/user-attachments/assets/9ecbf930-e2bd-4200-b662-c1e6d2cbb5d7

### DIFF
--- a/packages/touchpoint-ui/index.html
+++ b/packages/touchpoint-ui/index.html
@@ -102,47 +102,9 @@
         languageCode: "en-US",
       },
       windowSize: "full",
-      input: "text",
+      input: "voiceMini",
       initializeConversation: () => { }
     });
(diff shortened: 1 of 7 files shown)


jakub-nlx (approved):
At this point I'm wondering if having VoiceMini as a completely separate thing from Voice still makes sense?

peterszerzo:
> At this point I'm wondering if having VoiceMini as a completely separate thing from Voice still makes sense?

Yes I'm wondering the same thing, would hold off on a decision just a little bit because I'm anticipating all sorts of special needs coming to voice mini, but yeah we can keep this in mind.

PR #283, opened 2025-06-16
Onboarding Components

dtolb:


packages/touchpoint-components/README.md
```diff
@@ -0,0 +1,117 @@
...
+<html>
+<head>
+  <title>Touchpoint with Custom Components</title>
+</head>
+<body>
+  <!-- 1. Load Touchpoint UI -->
+  <script src="https://unpkg.com/@nlxai/touchpoint-ui/dist/touchpoint-ui.umd.js"></script>
+  
+  <!-- 2. Load Touchpoint Components -->
+  <script src="https://unpkg.com/@nlxai/touchpoint-components/dist/touchpoint-components.umd.js"></script>
```
dtolb, line 57: TBD

peterszerzo (changes requested):
I don't think publicly available touchpoint components should be responsible for `Museum...` concerns, this would mean that we are setting this package up to an infinitely growing collection of components for `Airport...`, `Hospical...`, `IceCreamShop...` and so on and so forth.

PR #297, opened 2025-06-19
Custom launch button

peterszerzo:
(feature still somewhat in discussion)

This feature allows the user to completely override the launch button with their own markup, with the option to get class definitions injected as props in a way that at least the positioning and the z-index stays intact.

CleanShot 2025-06-19 at 15 22 11@2x

https://github.com/user-attachments/assets/304bfd20-ebc2-4695-a712-a98ab1cc03c9


jakub-nlx:
To summarize my concerns (expressed on Slack) publicly:

1. This provides feature duplication with embedded mode (existing feature that isn't documented yet), as it is fairly trivial to make your own button with embedded mode (or `launchIcon: false` as @peterszerzo pointed out). So I think this is more of a documentation task to make it easy to copy-paste an easy to customize snippet rather than adding a dedicated and limited (see next point) feature just for this.
2. This will render inside [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM) meaning that **none** of the user's CSS styles are available. Normally this is a good thing, guaranteeing that Touchpoint and the host styles won't disrupt each other, but this makes it difficult for the user to style this custom button using their existing styling techniques (i.e. using their own design tokens, etc). Therefore, this solution makes making the button look the way you want to quite difficult for the user. Presumably the whole goal of this dedicated feature is to make doing that easy.

packages/touchpoint-ui/src/types.ts
```diff
@@ -177,6 +177,20 @@ export type InitializeConversation = (
...
 ) => void;
 
+/**
+ * Fully custom launch icon
+ */
+export type CustomLaunchButton = ComponentType<{
+  /**
+   * Class name injected into the component mainly to take care of positioning and z-index. Can be combined with more presentational and sizing-related class names.
+   */
+  className?: string;
```
jakub-nlx, line 187: Hm, I think we should provide more control here. For instance they might want to have the button in a different position (perhaps they have something important already in the bottom right corner), but still want to get correct z-index.

packages/touchpoint-ui/src/types.ts
```diff
@@ -207,7 +221,7 @@ export interface TouchpointConfiguration {
    *
    * When set to `false`, no launch button is shown at all. When not set or set to `true`, the default launch icon is rendered.
    */
-  launchIcon?: string | boolean;
+  launchIcon?: string | boolean | CustomLaunchButton;
```
jakub-nlx, line 224: Can we get rid of supporting the `string` here?

packages/touchpoint-ui/src/types.ts
peterszerzo: I had actually made a mistake here, z-index and placement is typically set on the container, fixing it here: eadbd0842fb171b32c7c813bc6f50a9b228b5f1c.

peterszerzo: Likely users will want simple `placement: "center-right"` stuff that just works.

jakub-nlx (approved):

PR #298, opened 2025-06-19
Bidi API improvements

jakub-nlx:
The idea here is to implement some of the high level API improvements suggested in https://github.com/nlxai/sdk/pull/286. 

These are all high level APIs implemented on top of the existing lower level APIs, so if you prefer the original style, that is still available for you. But getting a bidi voice plus touchpoint that can navigate and fill out forms is now as simple as:

```javascript
const touchpoint = await create({
    config: {
      applicationUrl: "YOUR_APPLICATION_URL",
      headers: {
        "nlx-api-key": "YOUR_API_KEY",
      },
      languageCode: "en-US",
      userId: crypto.randomUUID(), // Required for voice
    },
    input: "voiceMini", // Enables voice input with bidirectional support
    bidirectional: {},
  });
```

Doing that and having custom commands is also super straightforward:

```javascript
const touchpoint = await create({
    config: {
      applicationUrl: "YOUR_APPLICATION_URL",
      headers: {
        "nlx-api-key": "YOUR_API_KEY",
      },
      languageCode: "en-US",
      userId: crypto.randomUUID(), // Required for voice
    },
    input: "voiceMini", // Enables voice input with bidirectional support
    bidirectional: {
       custom: (action, payload) => {
           console.log(`Hey ${action} gave me`, payload);
       }
    },
  });
```

Opting out of these default behaviors and using the lower level API is as simple as:

```javascript
const touchpoint = await create({
    config: {
      applicationUrl: "YOUR_APPLICATION_URL",
      headers: {
        "nlx-api-key": "YOUR_API_KEY",
      },
      languageCode: "en-US",
      userId: crypto.randomUUID(), // Required for voice
    },
    input: "voiceMini", // Enables voice input with bidirectional support
    bidirectional: {
       automaticContext: false
    },
  });
```

---

On the implementation side, we use a mutation observer to gather context and look for links inside navs and footers for destinations. We re-send the context whenever this gathering changes. The default implementations of the commands are more or less what we have suggested in the documentation.

---

This works (with just the barest config):

https://github.com/user-attachments/assets/bd0aae25-f30b-499a-a7d5-64d0336993cb

(Screen recording doesn't record bot responses)

---

I added some sample changes to the documentation, but I think @dtolb will need to fine tune that, it's at best a rough draft.

### DIFF
--- a/packages/touchpoint-ui/src/index.tsx
+++ b/packages/touchpoint-ui/src/index.tsx
@@ -104,6 +104,7 @@ export {
   type CustomModalityComponent,
   type TouchpointConfiguration,
   type CustomLaunchButton,
+  type BidirectionalConfig,
 } from "./types";
 
 const defaultConversationId = (): string => {
--- a/packages/touchpoint-ui/src/types.ts
+++ b/packages/touchpoint-ui/src/types.ts
@@ -244,13 +244,38 @@ export type BidirectionalConfig =
       custom?: (action: string, payload: unknown) => void;
     }
   | {
+      /**
+       * Attempt to gather and send page context automatically. This will work well on semantically coded pages without too many custom form controls.
+       * This enables a number of automatic features.
+       *
+       * Defaults to `false`.
+       */
       automaticContext: false;
 
+      /**
+       * Navigation handler for bidirectional mode.
+       *
+       * If automatic context gathering is enabled, the default implementation will navigate to those pages using standard `window.location` APIs.
+       * @param action - The navigation action to perform.
+       * @param destination - The name of the destination to navigate to if `action` is `"page_custom"`.
+       */
       navigation?: (
         action: "page_next" | "page_previous" | "page_custom",
         destination?: string,
       ) => void;
+      /**
+       * A callback for filling out form fields in bidirectional mode.
+       *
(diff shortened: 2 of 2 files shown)


jakub-nlx:
This now includes resuming the conversation after navigation - this means that the bidi snippet we show in the docs actually works 😓 

@dtolb The other thing this PR does with docs consequences is that touchpoint now has defaults for both the `conversationId` and `userId` configuration variables that are both significantly better than just using `crypto.randomUUID()`. (They are still uuids, but conversationId is stored in session storage, user id in local storage, so both can be somewhat persistent.) As such, I would elide those from the snippets, since dumb copy-pasting would be somewhat counterproductive.

We should probably have a page that explains how to optimally use those 2 parameters:
- `conversationId` should be an ID of a conversation, that can persist through page reloads. Leaving the default implementation should be OK for most use cases, unless you wish to have a single conversation across multiple devices, in which case you will need to synchronize this value across devices.
- `userId` can be a unique user identifier. The default value will identify a unique user device and that not reliably, so tying this to a user login credential or similar will allow identifying a user reliably across devices. Also this identifier can then be used in analytics/conversation history for cross referencing with other systems and/or user support needs.

PR #300, opened 2025-06-20
Touchpoint embedded mode improvements

jakub-nlx:
Adds a number of improvements to touchpoint embedded mode so it can be successfully used by customers.

This will need some documentation @dtolb, albeit I don't think we need it everywhere, as I'd consider this to be a somewhat "advanced" feature.

What this PR does is:

1. Adds a way to define embedded touchpoints in HTML: 
     ```
     <nlx-touchpoint
        configuration='{
        "config": {
          "applicationUrl":
            "https://bots.dev.studio.nlx.ai/c/kQwAgKCx86vJ4r0PUSxYR/ON4zgwWnVofsz2NaiD38Z",
          "headers": {
            "nlx-api-key": "XDE2Th70VhBSlhz2YA"
          },
          "userId": "1234-5678",
          "languageCode": "en-US"
        },
        "input": "text"
      }'
      ></nlx-touchpoint>
     ```

     Of course this method is a little underpowered, since you can only define options that are expressible in JSON.

2. Cleans up the styling so you get a somewhat reasonably embedded touchpoint experience out of the box. Text and Voice touchpoints default to block elements 350px tall (a la SVG), Voice Mini defaults to being an intrinsically sized inline-block element. However, this styling is done with 0 specificity, so can easily be overriden by the user (and this is outside VDOM, so should be straightforward).
 
    However, there are probably further styling tweaks if we want to support small sized touchpoints, since the way touchpoint is layed out assumes a fairly large minimum size.

3. Removes close button in default embedded mode from touchpoint mini.

---

Some notes for documenting this feature:

1. All you need to do to get a touchpoint is have:
     1. The touchpoint script loaded (nothing needs to be imported out of it, it just needs to be loaded).
     2. Have a `<nlx-touchpoint>` HTML element in the DOM somewhere. You can either have it in your DOM, or create it with `document.createElement('nlx-touchpoint')` or even have something like React or Vue create it for you.
     3. Pass that element the configuration object. This can either be via the `configuration` **HTML Attribute** (which is a string, so the configuration needs to be serialized in JSON format), or via the `touchpointConfiguration` **JavaScript** property (see the example index file).
2. The following properties are irrelevant for embedded mode: `windowSize` and `launchIcon`.
3. The element instance has another property, `onClose`. By default this is `null` which means that the UI will not render any close buttons. If a function is passed, the close buttons will render and will cause the touchpoint to disappear if pressed. This behavior can be changed by something like `touchpointElem.onClose = (event) => { event.preventDefault(); doSomthingElse(instead); }`.
4. All other properties and setters on the element are marked internal and are not meant for customers to use (and can change without notice).

We already use embedded mode in the platform for onboarding, so that is an example use case.


peterszerzo (approved):
I'm curious: do the attribute and property names need to diverge? I have a feeling [some tools](https://svelte.dev/docs/svelte/custom-elements) figure out how to do the compilation with the same name.

peterszerzo (approved):

PR #303, opened 2025-06-20
Very Rough pass at bidi v+ in snippet

dtolb:
Incredibly rough cut at adding BiDi V+ setup snippet from onboarding to dev docs. 

Follow same user-experience: select the template from the "Try it live" and the snippet contains the copypasta code for user. 

Open Questions
* templateComponents maybe not the best name/param for such a bigger snippet change
* try it live bidi v+ with user snippet maybe doesn't make sense for this page... We'd basically be V+ driver on docs page.. Should we remove try-it-live for this scenario
* Given the number of config and other options to enable BiDi V+  I just created a net-new snippet. Probably a better way to solve.

PR only meant to start conversation

### DIFF
--- a/packages/website/src/content/02-01-touchpoint-ui-try-live.tsx
+++ b/packages/website/src/content/02-01-touchpoint-ui-try-live.tsx
@@ -35,7 +35,7 @@ export const snippetContent = ({
 }): string => {
   return `
 
-### Setup snippet${templateComponents === "museumComponents" ? ": Museum template" : ""}
+### Setup snippet${templateComponents === "museumComponents" ? ": Museum template" : templateComponents === "bidirectionalVoicePlus" ? ": Bidirectional Voice Plus template" : ""}
 
 \`\`\`touchpointui
 ${touchpointUiSetupSnippet({ config, theme, input, colorMode, templateComponents })}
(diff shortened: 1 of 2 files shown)


jakub-nlx:
Might be better when #298 lands, since then the bidi setup will be a lot more straightforward (effectively a one line change).

packages/website/src/content/02-01-touchpoint-ui-try-live.tsx
```diff
@@ -299,6 +317,10 @@ export const Content: FC<unknown> = () => {
                     value: "museumComponents",
                     label: "Museum",
                   },
+                  {
+                    value: "bidirectionalVoicePlus",
+                    label: "Bidirectional Voice Plus",
+                  },
```
jakub-nlx, line 323: I'd suggest perhaps making a checkbox that's "bidirectional" when the voiceMini input mode is selected?

packages/website/src/content/02-06-touchpoint-voice-plus.md
```diff
@@ -15,42 +14,29 @@
...
+  config: {
+    applicationUrl: "YOUR_APPLICATION_URL",
+    headers: {
+      "nlx-api-key": "YOUR_API_KEY",
     },
-    input: "voiceMini", // Enables voice input with bidirectional support
-    bidirectional: {},
-  };
+    languageCode: "en-US",
+    userId: crypto.randomUUID(), // Required for voice
```
jakub-nlx, line 31:
```suggestion
```

I'd remove this since it's no longer required.

packages/website/src/content/02-01-touchpoint-ui-try-live.tsx
dtolb: Good call! I updated so that it's a toggle that's only available if voiceMini is selected

PR #304, opened 2025-06-23
Remove terminate calls

peterszerzo:


peterszerzo:
> Do we want to remove the `terminateVoiceCall` api method as well?

@jakub-nlx yes, good call, just did!

PR #313, opened 2025-07-09
WIP: add transcription events

sam-trost:


packages/touchpoint-ui/src/voice.ts
```diff
@@ -196,6 +196,29 @@ export const useVoice = ({
...
+          if (
+            reader?.info?.attributes == null ||
+            reader.info.attributes["lk.transcribed_track_id"] == null
+          ) {
+            return;
+          }
+
+          const message = await reader.readAll();
+          if (reader.info.attributes["lk.transcribed_track_id"]) {
+            console.log(
```
sam-trost, line 211: for demonstration purposes...i'd assume we'd want to render these in an opinionated way as default and be able to disable it or (eventually) override it.

PR #316, opened 2025-07-17
Make sure fullscreen voice modalities are over the ripple

peterszerzo:


jakub-nlx (approved):
I wonder if we should re-do the layout to have less absolute positioning of everything.

peterszerzo:
> I wonder if we should re-do the layout to have less absolute positioning of everything.

I think in this case we need it by design, but the design will have to change for sure, @emily-nlx has more details.

PR #325, opened 2025-07-22
Fixes restoring text conversation

jakub-nlx:


jakub-nlx:
Does reset regenerate the internal conversation ID?

jakub-nlx:
Hm it does... but that probably won't get persisted correctly...

PR #333, opened 2025-07-24
Override for sending voice data

peterszerzo:
`handler.sendStuff` does not work over voice, there has to be an override in a way that they route to the right place and modalities can use the same code in handling user interactions.

### DIFF
--- a/packages/touchpoint-ui/src/voice.ts
+++ b/packages/touchpoint-ui/src/voice.ts
@@ -122,6 +122,7 @@ export const useVoice = ({
     setAudioElement(null);
     roomRef.current = null;
     setModalities([]);
+    handler.setRequestOverride(undefined);
     await room.disconnect();
   }, [setModalities, setAudioElement, handler]);
 


kai-nlx:
I pushed some changes here to use `publishData`, and I think it's working as expected now.

https://www.loom.com/share/dac074a12f8745d7958f6c6d160a70b1

packages/touchpoint-ui/src/voice.ts
```diff
@@ -201,6 +202,17 @@ export const useVoice = ({
         void disconnect();
       });
 
+      handler.setRequestOverride((req) => {
+        const encoder = new TextEncoder();
+        const encodedData = encoder.encode(JSON.stringify(req.request));
+        room.localParticipant
+          .publishData(encodedData)
+          .then(() => {})
```
sam-trost, line 210:
```suggestion
```

packages/touchpoint-ui/src/voice.ts
```diff
@@ -201,6 +202,16 @@ export const useVoice = ({
         void disconnect();
       });
 
+      handler.setRequestOverride((req) => {
+        const encoder = new TextEncoder();
+        const encodedData = encoder.encode(JSON.stringify(req.request));
+        room.localParticipant
+          .publishData(encodedData)
+          .catch((err) => {
+            console.error("Failed to publish data to LiveKit:", err);
```
sam-trost, line 211: @peterszerzo -- not sure how you handle errors in the SDK.  do we have a logging strategy or do we just swallow it 😱 

```suggestion
```

peterszerzo: @sam-trost the function inside `setRequestOverride` has another method called `append` which adds a response to the conversation transcript, where you can do the following: `append({ messages: [], modalities: {} })`, so I think we could signal voice errors with modalities (maybe we should set up a `DefaultError` modality?), which we can pipe through the voice modalities rendering system. I suppose the other option is to expand `append` so it can also do failure-type messages.

I don't think we quite have a solution fleshed out, just thinking out loud here.

sam-trost: i'll try to wrap this up this morning

peterszerzo: If we throw, whoever catches needs to trigger some UI feedback, and I don't think we quite have this fleshed out. I don't think it has anything to do with sync/async, the override just says 'any `handler.send...` method should run this function instead of what it would usually have done'

jakub-nlx: yeah but because those methods are sync, but here the error is asynchronous, the code that calls those methods has nowhere to catch the exception, right?

peterszerzo: Can't you just put a try-catch here: https://github.com/nlxai/sdk/blob/eda8c8b81015a588121cc762fec9c617d2b4e902/packages/core/src/index.ts#L1006-L1018?

jakub-nlx: I don't think so. That's not async, so the exception will happen after that code is already done.

PR #335, opened 2025-07-25
Set up response type enum

peterszerzo:


packages/core/src/index.ts
```diff
@@ -329,7 +347,7 @@ export interface FailureMessage {
   /**
    * The type of the response is `"application"` for application and `"user"` for user.
    */
-  type: "failure";
+  type: ResponseType.Failure;
```
jakub-nlx, line 350: The comment  here should also be changed. Although I think mentioning the literals when we have the enums is counter-productive.

jakub-nlx (changes requested):
There are also bits in touchpoint-ui that check these fields.

PR #338, opened 2025-08-13
Add way to define available custom actions and send them in

jakub-nlx:
Adds:

- `ToucpointInstance.setCustomBidirectionalCommands` for automatic context bidi mode. 
- `config.bidirectional.customizeAutomaticContext` option for customizing what the automatic context does. I think we've seen the use for this kind of function before - where the automatic mode does 90% of what you need, but there is some small feature you'd like to cover.
- Automatic context now sends the `uri` parameter automatically.

### DIFF
--- a/packages/touchpoint-ui/src/bidirectional/commandHandler.ts
+++ b/packages/touchpoint-ui/src/bidirectional/commandHandler.ts
@@ -1,16 +1,15 @@
 /* eslint-disable jsdoc/require-jsdoc */
 import type { ConversationHandler } from "@nlxai/core";
-import type { BidirectionalConfig } from "../types";
+import type { BidirectionalConfig } from "../interface";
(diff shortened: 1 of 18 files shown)


packages/touchpoint-ui/src/bidirectional/commandHandler.ts
```diff
@@ -90,7 +88,35 @@ export const commandHandler = (
         }
         break;
       case "custom":
```
peterszerzo, line 90: The handler should be called with all custom events, even the ones that have a handler defined in the new custom action definition method. This is to ensure legacy handlers work as expected in existing projects (even if a custom action will now be sent to the NLU with the new API, it is possible that the handler for it is in here).

peterszerzo (changes requested):

PR #342, opened 2025-08-25
Updates actions to new schema

jakub-nlx:
```ts
 touchpoint.setCustomBidirectionalCommands([{
    action: "Meal",
    description: "add a meal to your flight",
    schema: z.enum(["standard", "vegetarian", "vegan", "gluten-free"]),
    handler: (value) => {
      console.log("Meal option:", value);
    },
}]);
```

### DIFF
--- a/packages/touchpoint-ui/src/interface.ts
+++ b/packages/touchpoint-ui/src/interface.ts
@@ -224,7 +224,7 @@ export interface PageState {
   formElements: Record<string, Element>;
   /** Mapping from link element names to their URLs */
   links: Record<string, string>;
-  /** Mapping from custom commans to their handlers */
+  /** Mapping from custom commands to their handlers */
   customCommands: Map<string, (arg: any) => void>;
 }
 


packages/touchpoint-ui/package.json
```diff
@@ -39,6 +39,9 @@
     "react-textarea-autosize": "^8.5.6",
     "unist": "^0.0.1"
   },
+  "peerDependencies": {
```
peterszerzo, line 42: It's strange to me that zod is a peer dependency but React isn't.

packages/touchpoint-ui/index.html
```diff
@@ -90,6 +90,7 @@ <h1>Loreum ipsum dolor sit amet</h1>
     </main>
     <script type="module">
       import { create, html } from "/src/index.tsx";
+      import * as z from "zod/v4";
```
peterszerzo, line 93: How is this going to work for people who are not using a bundler? In the end it should be `import { z, create, html } from wherever;`, is there a way around that?

packages/touchpoint-ui/package.json
jakub-nlx: Well React really should be a peer dependency too...

packages/touchpoint-ui/index.html
jakub-nlx: Sort of up to them how they import it no? We can add another import line to our docs about this:

```html
<script type=module>
  import {create, html} from "some.cdn/@nlxai/touchpoint-ui"
  import * as z from "some-other.cdn/zod";
</script>
```

peterszerzo: In general my concern here is that we have to update the docs everywhere, depend on two cdn's, and have zod&React work inconsistently (one re-exported, the other imported from elsewhere). I don't think this is worth it.

packages/touchpoint-ui/package.json
jakub-nlx: Just because we made a mistake before shouldn't constrain us to make the same mistake again.

packages/touchpoint-ui/src/interface.ts
```diff
@@ -223,16 +224,8 @@ export interface PageState {
...
-  customCommands: Map<
-    string,
-    {
-      /** Available values, used for validation. */
-      values: any[];
-      /** Handler that will be called when the command is invoked */
-      handler: (arg: any) => void;
-    }
-  >;
+  /** Mapping from custom commans to their handlers */
```
peterszerzo, line 227:
```suggestion
  /** Mapping from custom commands to their handlers */
```

peterszerzo (changes requested):

PR #344, opened 2025-08-26
New Commands

dtolb:
## New commands

* `logout` - log out current user
* `whoami` - name of workspace and other information
* `hello` - hello world

## Others

* Added `consola` for output standardization
* Added "nlxSDK.ts" as an sdk-of-sorts for the management api ... see `whoami` command for how it's used
* Added types place-holder for SDK (should be able to supply these from the backend)

### DIFF
--- a/packages/cli/src/commands/hello.ts
+++ b/packages/cli/src/commands/hello.ts
@@ -1,0 +1,12 @@
+import { Command } from 'commander';
+import { consola } from 'consola';
+
+const HELLO_DESCRIPTION = 'Display a hello world message';
+
+export const helloCommand = new Command('hello')
+  .description(HELLO_DESCRIPTION)
+  .option('-n, --name <name>', 'name to greet', 'World')
+  .action((options) => {
+    consola.success('🚀 Hello from the NLX CLI!');
+    consola.info(`Hello, ${options.name}!`);
+  });
--- a/packages/cli/src/commands/logout.ts
+++ b/packages/cli/src/commands/logout.ts
@@ -1,0 +1,20 @@
+import { Command } from 'commander';
+import { consola } from 'consola';
+import { clearTokens } from '../commands/login';
+
+const LOGOUT_SUCCESS_MESSAGE = 'Logged out. Tokens cleared from storage.';
+const LOGOUT_DESCRIPTION = 'Clear stored authentication tokens';
+
+export const logoutCommand = new Command('logout')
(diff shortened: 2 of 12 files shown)


packages/cli/src/commands/hello.ts
```diff
@@ -0,0 +1,12 @@
+import { Command } from 'commander';
```
jakub-nlx, line 1: Let's get rid of this file.

packages/cli/src/commands/logout.ts
```diff
@@ -0,0 +1,20 @@
...
+import { clearTokens } from '../commands/login';
+
+const LOGOUT_SUCCESS_MESSAGE = 'Logged out. Tokens cleared from storage.';
+const LOGOUT_DESCRIPTION = 'Clear stored authentication tokens';
+
+export const logoutCommand = new Command('logout')
+  .description(LOGOUT_DESCRIPTION)
+  .action(async () => {
+    try {
+      await clearTokens();
```
jakub-nlx, line 12: Is there no server side thing we need to do to invalidate the tokens?

packages/cli/src/commands/whoami.ts
```diff
@@ -0,0 +1,45 @@
+import { Command } from 'commander';
+import { consola } from 'consola';
+import { team as teamApi, bots as botsApi } from '../utils/nlxSDK';
+import type { BotsListResponse } from '../utils/types/api';
+import { ensureToken } from '../commands/login';
+
+const WHOAMI_DESCRIPTION = 'Show current team, role, and available bots count (requires login)';
```
jakub-nlx, line 7: We don't use the `bots` language anymore.

packages/cli/src/utils/types/api.ts
```diff
@@ -0,0 +1,40 @@
+export type ApiRequestBase = {
+    [key: string]: unknown;
+  };
```
jakub-nlx, line 3: This won't really do anything, since this is how typescript behaves by default.

packages/cli/src/utils/types/api.ts
```diff
@@ -0,0 +1,40 @@
+export type ApiRequestBase = {
```
jakub-nlx, line 1: Not sure what the benefit of these types is... we don't have a way to guarantee they're accurate unless we bother with schema validation with something like Zod.

PR #348, opened 2025-09-01
Fixes nlxActiveVoiceConversation being set after closing the widget

jakub-nlx:


packages/touchpoint-ui/src/App.tsx
```diff
@@ -99,17 +99,20 @@ const App = forwardRef<AppRef, Props>((props, ref) => {
...
   const input = props.input ?? "text";
 
+  const hangUp = useCallback(() => {
+    sessionStorage.removeItem("nlxActiveVoiceConversationId");
+    if (input === "voice" || input === "voiceMini") {
+      conversationInitialized.current = false;
+      handler.reset({ clearResponses: true });
+      sessionStorage.removeItem("nlxConversationId");
+    }
+  }, []);
```
peterszerzo, line 109: Add `handler` and `input` here as a hook argument.

packages/touchpoint-ui/src/App.tsx
```diff
@@ -171,11 +181,9 @@ const App = forwardRef<AppRef, Props>((props, ref) => {
...
     }
-    if (props.config.conversationId != null)
-      sessionStorage.setItem(
-        "nlxActiveVoiceConversationId",
-        props.config.conversationId,
-      );
+    const newConversationId = handler.currentConversationId();
+    if (newConversationId != null)
+      sessionStorage.setItem("nlxActiveVoiceConversationId", newConversationId);
   }, [props.initializeConversation, props.initialContext, handler, isExpanded]);
```
peterszerzo, line 187: Add `hangUp` here as a hook argument.

peterszerzo (changes requested):

packages/touchpoint-ui/src/App.tsx
```diff
@@ -99,23 +99,26 @@ const App = forwardRef<AppRef, Props>((props, ref) => {
...
 
+  const hangUp = useCallback(() => {
+    sessionStorage.removeItem("nlxActiveVoiceConversationId");
+    if (input === "voice" || input === "voiceMini") {
+      conversationInitialized.current = false;
+      handler.reset({ clearResponses: true });
+      sessionStorage.removeItem("nlxConversationId");
+    }
+  }, [handler, input]);
+  const _onClose = props.onClose;
```
peterszerzo, line 110: Is this set up here just so the hook rule can do its job more easily? This would read terribly if every destructuring we did introduced these underscore variables everywhere.

peterszerzo (changes requested):

peterszerzo (approved):

PR #355, opened 2025-09-10
Sets the rive color to --accent

jakub-nlx:
https://github.com/user-attachments/assets/201921ad-f60f-4a25-a14f-3790d8d8d6dd


peterszerzo (approved):
I love this implementation so much ❤️

PR #361, opened 2025-09-16
Make animation respond to voice level

jakub-nlx:
This may be actually too boring, but since I had it there, we may as well have the code available.

### DIFF
--- a/packages/touchpoint-ui/src/components/RiveAnimation.tsx
+++ b/packages/touchpoint-ui/src/components/RiveAnimation.tsx
@@ -25,9 +25,13 @@ function resolveCssVariable(
   ];
 }
 
-export const RiveAnimation: FC<{ restored: boolean }> = ({ restored }) => {
+export const RiveAnimation: FC<{ restored: boolean; volumeLevel: number }> = ({
+  restored,
+  volumeLevel,
+}) => {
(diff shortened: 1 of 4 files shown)


packages/touchpoint-ui/src/components/RiveAnimation.tsx
```diff
@@ -25,9 +25,13 @@ function resolveCssVariable(
...
 }
 
-export const RiveAnimation: FC<{ restored: boolean }> = ({ restored }) => {
+export const RiveAnimation: FC<{ restored: boolean; volumeLevel: number }> = ({
+  restored,
+  volumeLevel,
+}) => {
   const samplerRef = useRef<HTMLCanvasElement | null>(null);
   const ref = useRef<HTMLCanvasElement | null>(null);
+  const voiceInputRef = useRef<{ value: number }>({ value: 0 });
```
peterszerzo, line 34: I don't get what's happening with this ref, it's just defined, written to, and not read anywhere. I also don't see the voiceInput rive variable being written to.

peterszerzo (changes requested):

packages/touchpoint-ui/src/components/RiveAnimation.tsx
peterszerzo: Oooooh I get it, clever 😎

PR #362, opened 2025-09-16
Call event handler without payload

dtolb:
Checks the handler function parameter expectations

* If no payload parameter is expected, call the handler if null
* If payload parameter is expected, don't call, but warn console we ended up here

### DIFF
--- a/packages/touchpoint-ui/src/bidirectional/commandHandler.ts
+++ b/packages/touchpoint-ui/src/bidirectional/commandHandler.ts
@@ -96,12 +96,25 @@ export const commandHandler = (
       case "custom":
         if (pageState.current.customCommands.has(event.action as string)) {
           // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          const handler = pageState.current.customCommands.get(
+          const customHandler = pageState.current.customCommands.get(
             event.action as string,
           )!;
 
-          if (event.payload != null) {
-            handler(event.payload);
+          const expectsArgs =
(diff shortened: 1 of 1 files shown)


packages/touchpoint-ui/src/bidirectional/commandHandler.ts
```diff
@@ -96,12 +96,25 @@ export const commandHandler = (
...
           // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          const handler = pageState.current.customCommands.get(
+          const customHandler = pageState.current.customCommands.get(
             event.action as string,
           )!;
 
-          if (event.payload != null) {
-            handler(event.payload);
+          const expectsArgs =
+            typeof customHandler === "function" && customHandler.length > 0;
```
jakub-nlx, line 104: I think `Function.prototype.length` is fairly unreliable especially when we don't control the callsite.

I would generally recommend users use something like Zod for defining the schema and then re-use the same schema to validate and type check the argument before further processing... we used to do this ourselves, but we can't quite afford the dependency.

jakub-nlx (changes requested):

peterszerzo:
Why not just this? I suppose you could fish out the schema again and do a runtime json-schema check if it exists, but I don't see a reason why payloadless commands are illegitimate.

CleanShot 2025-09-17 at 08 56 11@2x

dtolb:
Works for me

PR #365, opened 2025-09-24
Refactors voice into core

jakub-nlx:
> Why is this a draft? The precise API of this module is experimented with on a few call sites, adding features like transcriptions and clearer data modeling. Once those implementations mature, we will move them into core.

This will make it easier for us to build voice experiences outside touchpoint (such as the new testing widget).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22344,6 +22344,7 @@
       "license": "MIT",
       "dependencies": {
         "isomorphic-fetch": "^3.0.0",
+        "livekit-client": "^2.9",
         "ramda": "^0.29.1",
         "reconnecting-websocket": "^4.4.0",
         "uuid": "^9.0.1"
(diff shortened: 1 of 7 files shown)


peterszerzo (changes requested):
Oh interesting! Two thoughts:
1. I'm seeing two things happening here: changing the API so it is generic and relies on hooks less (which is fantastic ✨), and moving it all to core. I would like to be prudent and do roll out just the former while keeping it in touchpoint-ui.
2. core must work in node. Could we check if this would cause any hiccups? I have a feeling`livekit-client` might have some web dependencies, and those can lead to a straight up runtime error when you simply want to instantiate a regular conversation client in a lambda.

peterszerzo:
> Tested this with node and standard functions work fine (i.e. `subscribe`, `sendWelcomeFlow`, etc).
> 
> `initiateVoiceConnection` rejects the promise with `new Error("Failed to start audio")`, which seems reasonable in a Node context?
> 
> On keeping this within touchpoint: I don't see how that achieves much in either goal:
> 
> 1. It doesn't simplify forking touchpoint, since you still have to deal with live kit and the giant ball of complexity that represents.
> 2. It doesn't allow you to build stuff without touchpoint. I don't even really know how we would expose this in a coherent way from the touchpoint package?

My only goal here is to be cautious and only extract things over to core once they have proven successful in existing voice packages for a while.

PR #367, opened 2025-09-30
Warn if there is no handler defined on an action

peterszerzo:


packages/touchpoint-ui/src/bidirectional/commandHandler.ts
```diff
@@ -101,6 +101,10 @@ export const commandHandler = (
           )!;
           handler(event.payload);
         }
+        // eslint-disable-next-line no-console
+        console.warn(
+          `No command handler was defined for the ${event.action} action.`,
```
jakub-nlx, line 106:
```suggestion
        debug(
          `No custom command handler was defined for the %o action.\n\n%cTip: Set up a handler with \nsetCustomBidirectionalCommands([{ action: "${event.action}", handler() { }}])`, event.action, "font-style: italic; font-size: 90%"
```

Screenshot 2025-09-30 at 11 25 15

Something like that?

jakub-nlx (approved):

PR #415, opened 2026-01-19
Tooltips

jakub-nlx:
https://github.com/user-attachments/assets/420b7e80-5c80-4d6e-97f0-9cb9404a9eb2


peterszerzo (changes requested):
Check paddings, colors and border radii (the latter is definitely too large).

peterszerzo (changes requested):
Looks like the button disabled state broke between this branch and main

Image

PR #417, opened 2026-01-19
Remove website

peterszerzo:
- remove `website` package.
- upgrade `lerna` (takes care of some of the vulnerabilities).


jakub-nlx:
This might also unblock us from upgrading React and associated deps in Touchpoint, right?

Issue #424, opened 2026-01-27
Use `dompurify` for markdown


peterszerzo:
We should probably adopt our SafeMarkdown component from the main app. This is potentially an XSS vector (although some details of how one might exploit this remain to be seen).

_Originally posted by @jakub-nlx in https://github.com/nlxai/sdk/pull/423#discussion_r2732361870_

PR #428, opened 2026-02-04
Sanitize markdown rendered output

peterszerzo:
Fixes https://github.com/nlxai/sdk/issues/424.

### DIFF
--- a/packages/touchpoint-ui/src/components/Input.tsx
+++ b/packages/touchpoint-ui/src/components/Input.tsx
@@ -15,7 +15,6 @@ import {
   ResponseType,
 } from "@nlxai/core";
 import { clsx } from "clsx";
-import DOMPurify from "dompurify";
 
 import { useCopy } from "../utils/useCopy";
 import { IconButton } from "./ui/IconButton";


packages/touchpoint-ui/src/components/ui/TextButton.tsx
```diff
@@ -68,7 +70,7 @@ export const TextButton: FC<TextButtonProps> = ({
         className,
       )}
     >
-      {label}
+      {DOMPurify.sanitize(label)}
```
jakub-nlx: I don't think this is necessary. React will not pass through HTML here as it will use a safe API to insert text here.

packages/touchpoint-ui/src/components/FullscreenVoice.tsx
```diff
@@ -139,7 +140,9 @@ export const VoiceModalities: FC<{
                     <div
                       className={clsx("space-y-6 markdown")}
                       dangerouslySetInnerHTML={{
-                        __html: marked(message.text, { async: false }),
+                        __html: DOMPurify.sanitize(
+                          marked(message.text, { async: false }),
+                        ),
```
jakub-nlx: Let's make a `SafeMarkdown` component that does all of this and use it instead of leaving this to the callers.

vlad31 (approved):

packages/touchpoint-ui/src/components/Input.tsx
```diff
@@ -102,13 +103,21 @@ export const Input: FC<InputProps> = ({
 
     setIsWaiting(true);
 
+    // Disallow all user-input HTML
+    // TODO: also add this logic to the NLU
+    const sanitizedInputValue = DOMPurify.sanitize(inputValue, {
+      ALLOWED_TAGS: [],
+    });
```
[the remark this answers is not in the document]
peterszerzo: Good argument, will remove.

PR #443, opened 2026-03-13
use adoptedstylesheets

ricardo-nlx:
# adoptedStyleSheets migration

## The problem

Tailwind v4 uses [`@property`](https://developer.mozilla.org/en-US/docs/Web/CSS/@property) to register CSS custom properties with explicit types and initial values — for example `--tw-translate-x` with `initial-value: 0`. **`@property` declarations are silently ignored inside a shadow root.** They must be registered at the document level.

Previously the compiled Tailwind CSS was injected as a `<style>` tag inside the closed shadow root, which broke all `@property`-dependent utilities. The codebase worked around this with redundant zero-value translate classes:

```tsx
"md:transform md:-translate-x-full md:translate-y-0"  // translate-y-0 is the workaround
"transform translate-x-0 -translate-y-full ..."        // translate-x-0 is the workaround
```

This isn't just an aesthetic/verbosity problem with the current codebase. We spent a couple of hours debugging this already, so it's a mainteinance concern.

## The fix

Share a single `CSSStyleSheet` across both the shadow root and the document.

```ts
const sheet = new CSSStyleSheet();
sheet.replaceSync(cssRaw);

// in #render():
this.#shadowRoot.adoptedStyleSheets = [sheet];
if (!document.adoptedStyleSheets.includes(sheet)) {
  document.adoptedStyleSheets = [...document.adoptedStyleSheets, sheet];
}
```

The redundant zero-value workaround classes were also removed from `Header.tsx`, `VoiceMini.tsx`, and `FullscreenVoice.tsx`.

## Visual comparison (`windowSize: "half"`)

The header controls should float outside the chat panel to the left via `md:-translate-x-full`.

**Before** — buttons stuck inside the panel, overlapping content:

adopted-stylesheets-before

**After** — buttons correctly positioned in the overlay area:
adopted-stylesheets-after

## Notes on `@property` collisions

- **Tailwind v3** does not use `@property` at all 
- **Tailwind v4 on the same page**: the only realistic collision scenario. Harmless in practice: all Tailwind v4 installs generate identical registrations for the same property names.
- **Other frameworks** — very very unlikely to register `--tw-*` properties. If it ever became a concern, we could do css post-processing and prefix our properties with `nlx:`

### DIFF
--- a/packages/touchpoint-ui/__tests__/adoptPropertyDeclarations.test.ts
+++ b/packages/touchpoint-ui/__tests__/adoptPropertyDeclarations.test.ts
@@ -1,0 +1,27 @@
+import { describe, it, expect } from "vitest";
+import { extractPropertyCSS } from "../src/adoptPropertyDeclarations";
+
+const sampleCSS = `
+  .foo { color: red; }
+  @property --tw-translate-x { syntax: "*"; inherits: false; initial-value: 0 }
+  .bar { display: flex; }
+  @property --glow-opacity { syntax: "<percentage>"; inherits: false; initial-value: 0% }
+`;
+
+describe("extractPropertyCSS", () => {
+  it("extracts only @property declarations from the CSS", () => {
+    const result = extractPropertyCSS(sampleCSS);
+    expect(result).toContain("@property --tw-translate-x");
+    expect(result).toContain("@property --glow-opacity");
+    expect(result).not.toContain(".foo");
+    expect(result).not.toContain(".bar");
+  });
+
+  it("returns null when no @property declarations are present", () => {
+    expect(extractPropertyCSS(".foo { color: red; }")).toBeNull();
+  });
+
+  it("returns null for an empty string", () => {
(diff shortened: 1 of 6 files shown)


jakub-nlx (dismissed):

jakub-nlx (commented):
I'm not 100% convinced this will not clobber some styles if they are also using Tailwind... since this will load our tailwind and their tailwind into the same namespace no?

(I get the `@properties` are the same, but presumably some of the classes can have the same names but different definitions...)

peterszerzo (changes requested):
To @jakub-nlx 's point above, I'm not convinced this is safe. Let's leave it around as a concept in any case.

jakub-nlx (approved):
I think this is a reasonable solution. Thoughts @peterszerzo ?

peterszerzo:
@jakub-nlx let's discuss at a later point whether this is ready to merge, until then I moved to drafts.

PR #473, opened 2026-06-22
+ Amazon Connect Chat adapter for Touchpoint UI

ndrppnc:
Introduces @nlxai/connect-chat-adapter that bridges Amazon Connect Chat (via amazon-connect-chatjs) into the Touchpoint UI through the shared ConversationHandler interface. Supports text, choices, interactive messages, modalities (via application/json), and session reset with fresh contact creation. Also adds a demo page (connect-chat.html) with loading animation on initial connect.

### DIFF
--- a/packages/connect-chat-adapter/package.json
+++ b/packages/connect-chat-adapter/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@nlxai/connect-chat-adapter",
+  "version": "0.1.0",
+  "description": "Amazon Connect Chat adapter for Touchpoint UI",
+  "type": "module",
+  "main": "lib/index.cjs",
+  "module": "lib/index.esm.js",
+  "browser": "lib/index.umd.js",
+  "types": "lib/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
+      "import": "./lib/index.esm.js",
+      "require": "./lib/index.cjs"
+    }
+  },
+  "scripts": {
+    "build": "rm -rf lib && rollup -c --configPlugin typescript --configImportAttributesKey with",
+    "lint:check": "eslint src/",
+    "lint": "eslint src/ --fix",
+    "prepublish": "npm run build",
+    "tsc": "tsc"
+  },
+  "author": "NLX <engineering@nlx.ai>",
+  "license": "MIT",
+  "devDependencies": {
+    "@nlxai/eslint-config": "*",
+    "@nlxai/rollup-config": "*",
+    "@rollup/plugin-commonjs": "^25.0.7",
+    "@rollup/plugin-json": "^6.1.0",
+    "@rollup/plugin-node-resolve": "^15.2.3",
+    "@rollup/plugin-replace": "^5.0.5",
+    "@rollup/plugin-terser": "^0.4.4",
+    "@rollup/plugin-typescript": "^11.1.6",
+    "@types/node": "^24.10.1",
+    "rollup": "^4.59.0",
+    "typescript": "^5.5.4"
+  },
+  "dependencies": {
+    "@nlxai/core": "^1.2.6"
+  },
+  "peerDependencies": {
+    "amazon-connect-chatjs": "^2.0.0"
+  },
+  "peerDependenciesMeta": {
+    "amazon-connect-chatjs": {
+      "optional": false
+    }
+  },
+  "publishConfig": {
+    "access": "public"
+  }
+}
--- a/packages/connect-chat-adapter/package.json
+++ b/packages/connect-chat-adapter/package.json
@@ -37,10 +37,8 @@
     "typescript": "^5.5.4"
   },
   "dependencies": {
-    "@nlxai/core": "^1.2.6"
-  },
-  "peerDependencies": {
-    "amazon-connect-chatjs": "^2.0.0"
+    "@nlxai/core": "^1.2.6",
+    "amazon-connect-chatjs": "^2.3.2"
   },
   "peerDependenciesMeta": {
     "amazon-connect-chatjs": {
--- a/packages/connect-chat-adapter/package.json
+++ b/packages/connect-chat-adapter/package.json
@@ -19,7 +19,11 @@
     "lint:check": "eslint src/",
     "lint": "eslint src/ --fix",
     "prepublish": "npm run build",
-    "tsc": "tsc"
+    "test": "typedoc --emit none",
+    "tsc": "tsc",
+    "update-readme:docs": "rm -rf docs/ && typedoc",
+    "update-readme:merge": "../../scripts/transclude-markdown.js",
+    "update-readme": "npm run update-readme:docs && npm run update-readme:merge"
   },
   "author": "NLX <engineering@nlx.ai>",
   "license": "MIT",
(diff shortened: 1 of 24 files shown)


packages/connect-chat-adapter/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+    contactAttributes?: Record<string, string>;
+    /**
+     * Content types the chat participant supports receiving.
+     * Defaults to text/plain, text/markdown, application/json, and interactive messages.
+     */
+    supportedMessagingContentTypes?: string[];
+  };
+  /**
+   * Pre-obtained chat details from a prior StartChatContact call.
+   * Mutually exclusive with `startChatEndpoint`.
```
jakub-nlx: Worth encoding in the type?

i.e.

```ts
export type ConnectChatConfig = 
{ 
   startChatEndpoint?: string;
   startChatParams?: {
    instanceId?: string;
    contactFlowId?: string;
    participantDisplayName?: string;
    contactAttributes?: Record<string, string>;
    supportedMessagingContentTypes?: string[];
  }
} | {
 chatDetails?: {
    contactId: string;
    participantId: string;
    participantToken: string;
  };
} & {
  region?: string;
  languageCode?: LanguageCode;
  globalConfig?: Record<string, unknown>;
};

File: packages/connect-chat-adapter/src/index.ts [hunk 2]
```diff
@@ -0,0 +1,611 @@
...
+}
+
+/**
+ * Creates a ConversationHandler backed by Amazon Connect Chat via amazon-connect-chatjs.
+ *
+ * This adapter conforms to the same interface used by \@nlxai/core's `createConversation`,
+ * so it can be passed directly to Touchpoint UI via the `conversationHandler` prop.
+ *
+ * Prerequisites:
+ * - `amazon-connect-chatjs` must be loaded before calling this function
```
jakub-nlx: Any reason we are not calling `import "amazon-connect-chatjs";` in our code if it is a pre-requisite?

packages/connect-chat-adapter/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+ */
+export interface ConnectChatConversationHandler extends ConversationHandler {
+  /**
+   * Emit a custom interim message to display in Touchpoint's loading state.
+   * Pass `undefined` to clear it.
+   */
+  emitInterimMessage: (message?: string) => void;
+}
+
+export function createConnectChatConversation(
```
jakub-nlx: Should have an API doc

packages/connect-chat-adapter/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+  };
+
+  const fetchChatDetails = async (): Promise<ResolvedChatDetails> => {
+    if (config.chatDetails != null) {
+      return config.chatDetails;
+    }
+    if (config.startChatEndpoint == null) {
+      throw new Error(
+        "@nlxai/connect-chat-adapter: Either 'chatDetails' or 'startChatEndpoint' must be provided.",
+      );
```
jakub-nlx: I'd make our own Error sublass that prepends the `"@nlxai/connect-chat-adapter"`.

packages/connect-chat-adapter/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+      body.ContactFlowId = config.startChatParams.contactFlowId;
+    }
+    if (config.startChatParams?.participantDisplayName != null) {
+      body.ParticipantDetails = {
+        DisplayName: config.startChatParams.participantDisplayName,
+      };
+    }
+    if (config.startChatParams?.contactAttributes != null) {
+      body.Attributes = config.startChatParams.contactAttributes;
+    }
```
jakub-nlx: Perhaps we should match connect terminology/casing exactly in our public API? (or does their JS also change these names?)

In fact if they publish their TS definitions, it might be even better to just import them and let the users pass these through rather then specifying a specific subset.

packages/connect-chat-adapter/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+  const unsubscribe = (subscriber: Subscriber): void => {
+    subscribers = subscribers.filter((fn) => fn !== subscriber);
+  };
+
+  const handler: ConnectChatConversationHandler = {
+    sendText,
+    sendChoice,
+
+    sendSlots: (_slots: SlotsRecordOrArray, _context?: Context): void => {
+      // No-op: Connect Chat does not have a slots concept
```
jakub-nlx: Perhaps throw then?

packages/connect-chat-adapter/src/index.ts
```diff
@@ -0,0 +1,611 @@
...
+    sendWelcomeFlow: (_context?: Context): void => {
+      // Connect Chat triggers the contact flow automatically on connection
+    },
+
+    sendWelcomeIntent: (_context?: Context): void => {
+      // Deprecated, no-op
+    },
+
+    sendFlow: (_flowId: string, _context?: Context): void => {
+      // No-op: Connect Chat does not support arbitrary flow triggering from client
```
jakub-nlx: Also throw.

packages/connect-chat-adapter/package.json
```diff
@@ -0,0 +1,53 @@
...
+      "import": "./lib/index.esm.js",
+      "require": "./lib/index.cjs"
+    }
+  },
+  "scripts": {
+    "build": "rm -rf lib && rollup -c --configPlugin typescript --configImportAttributesKey with",
+    "lint:check": "eslint src/",
+    "lint": "eslint src/ --fix",
+    "prepublish": "npm run build",
+    "tsc": "tsc"
```
jakub-nlx: Should copy the README generation script from the other packages so we can get nice API docs.

jakub-nlx (changes requested):

packages/connect-chat-adapter/src/index.ts
peterszerzo: Oh yes right on, we must import it, otherwise it's extra stuff we need to remember in the docs and we get into all sorts of content security debates.

packages/connect-chat-adapter/src/index.ts
peterszerzo: Resolved in the follow-up PR.

packages/connect-chat-adapter/src/index.ts
peterszerzo: We will no longer need this per #475

packages/connect-chat-adapter/src/index.ts
peterszerzo: We can't really throw here, the end user might legitimately call these and end up with an uncaught exception. We will warn though.

packages/connect-chat-adapter/src/index.ts
peterszerzo: See above.

packages/connect-chat-adapter/package.json
peterszerzo: On it!

packages/connect-chat-adapter/src/index.ts
peterszerzo: #475 separates the details retrieval and the conversation initialization, so `details` will be mandatory here, and the former two will be mandatory for the fetching.

packages/connect-chat-adapter/package.json
jakub-nlx: (And also run it and commit the results)

jakub-nlx (dismissed):

packages/connect-chat-adapter/package.json
peterszerzo: Yes that's done as well

jakub-nlx (approved):

pszerzo-amazon (approved):

PR #477, opened 2026-06-24
Update codeowners

pszerzo-amazon:


jakub-nlx (approved):
I guess you are changing accounts?

pszerzo-amazon:
> I guess you are changing accounts?

Yeah I want to have a global `.gitconfig` that always has the same email address.

PR #479, opened 2026-06-26
Make core work for sandbox

pszerzo-amazon:
When using core with the `.experimental.completeApplicationUrl` feature flag, forego URL parsing altogether.


jakub-nlx (approved):
Earlier there was a problem with my `npm link` setup. Verified this actually works.

PR #480, opened 2026-06-26
Update command payload

choso:


peterszerzo (approved):

pszerzo-amazon:
Thanks for the update @choso - I went ahead and added a detailed upfront type definition that allows us to remove all the type casting.

We'll wait for @jakub-nlx to take a look at it and then we're good to go.

jakub-nlx (approved):
Do we want to validate the event at the network boundary?

pszerzo-amazon:
> Do we want to validate the event at the network boundary?

I'm all for it as long as we stay on the lenient side and don't throw errors.

